PR #40, opened 2020-06-02
[REVIEW] Add spdlog to rapids-build-env

shwina:
Required when building RMM Cython.

PR introducing this requirement: https://github.com/rapidsai/rmm/pull/363


raydouglass:
This is needed and already done in `branch-0.15`

PR #45, opened 2020-06-05
[REVIEW] Update pyarrow/arrow-cpp versions to 0.17.1, add arrow-cpp-proc CUDA package

trxcllnt:


conda/recipes/release-versions.yaml
```diff
@@ -24,7 +24,7 @@ build_stack_version:
 
 # Shared versions across meta-pkgs
 arrow_version:
-  - '=0.15.0'
+  - '=0.17.1'
```
kkraus14, line 27: Note that RAPIDS 0.14 still requires arrow 0.15.0. @dillon-cullinan should this be updated to account for RAPIDS 0.15 release or stay in accordance with RAPIDS 0.14?

mike-wendt: @kkraus14 we have a `branch-0.14` that we use for release 0.14 so update both here as this will serve as release 0.15

kkraus14:
@trxcllnt we need to bump the `librdkafka` version as well or this won't solve

kkraus14:
@mike-wendt bumped `librdkafka`

PR #61, opened 2020-06-22
Bump pandas to 1.0+

brandon-b-miller:


kkraus14:
@raydouglass should we bump the nightly version as well?

mike-wendt:
> 
> 
> @raydouglass should we bump the release version as well?
not at the moment - we'll handle that at burndown

PR #68, opened 2020-07-08
ENH Consolidate the two versions files into one

mike-wendt:
This is to adapt the previous workflow that did not include versioned branches. With versioned branches we do not need a nightly and a release version file. Instead we can use one file, and reduce the testing. In addition to the consolidation of these files, the PR test jobs will need to be updated along with the documentation.


mike-wendt:
Change in direction will open new PR

PR #72, opened 2020-07-13
Require `pickle5` (on older Python versions)

jakirkham:
For older versions of Python, make sure the `pickle5` backport package is available. This way we can still leverage the features of pickle protocol 5 (namely out-of-band buffer serialization) on older Python versions.

https://docs.python.org/3/library/pickle.html#pickle-oob
https://www.python.org/dev/peps/pep-0574/
https://github.com/pitrou/pickle5-backport


jakirkham:
cc @kkraus14 @dantegd @hcho3 @quasiben

PR #74, opened 2020-07-14
[REVIEW] Pinning black, flake8 and isort

rgsl888prabhu:
black, flake8 and isort are not controlled by their versions. And sometimes these break and block CIs. So planning to control the version of these to ensure smooth CI operation.

It is open for review, but as this will affect all the RAPIDS projects, wanted to check with PICs of those projects if they have any concerns or any other approach to handle this issue.

### DIFF
--- a/conda/recipes/rapids-build-env/meta.yaml
+++ b/conda/recipes/rapids-build-env/meta.yaml
@@ -41,7 +41,7 @@ requirements:
     - async_generator
     - autoconf
     - automake
-    - black
+    - black ={{ black_version }}
     - conda-forge::blas{{ blas_version }}
     - boost
     - boost-cpp {{ boost_cpp_version }}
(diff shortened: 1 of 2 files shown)


mike-wendt:
@rgsl888prabhu looks like there are conflicts with `branch-0.15`, can you resolve them so we can review?

I'm pulling in @jakirkham @jrhemstad @kkraus14 for reviewers to this as well

jakirkham:
Would suggest we relax the pinnings a bit and pin to major versions instead.

rgsl888prabhu:
> Would suggest we relax the pinnings a bit and pin to major versions instead.

If we pin to major version, pkg will still get update for any new minor versions, which might create similar problems we have been facing.

jakirkham:
The problem we saw with `black` was due to it upgrading to `5.0`. Major versions can and do break compatibility. What I'm proposing takes that into account.

kkraus14:
FWIW we've also consistently seen breakages with flake8 minor version releases

jakirkham:
Could you please share some examples?

kkraus14:
> Could you please share some examples?

Going from 3.7 --> 3.8: https://github.com/rapidsai/cudf/pull/5185
Don't remember the specific version change: https://github.com/rapidsai/cudf/pull/821/files

jakirkham:
Thanks! If we want to pin `flake8` to minor version and let the patch version float, that seems fine.

PR #82, opened 2020-07-27
[REVIEW] Add custreamz to rapids

dillon-cullinan:
Adding the custreamz meta-package to the rapids meta-package. custreamz is available in the devel containers since it installs the streamz dependency via `rapids-build-env`, but unavailable in base/runtime since they use `rapids` instead of `rapids-build-env`.


kkraus14:
Fixed by #86

PR #86, opened 2020-07-29
Added python-confluent-kafka dependency to conda build-env recipe

### LINKED ISSUE #85: Add python-confluent-kafka to rapids-build-env conda recipe
Current work on https://github.com/rapidsai/cudf/pull/5782 requires that `python-confluent-kafka` be present in the conda build environment.

jdye64:
Added `python-confluent-kafka` to conda recipe and also added `custreamz` to the rapids recipe since that is also needed now.

This closes #85

### DIFF
--- a/conda/recipes/versions.yaml
+++ b/conda/recipes/versions.yaml
@@ -91,6 +91,8 @@ panel_version:
   - '>=0.9.*'
 pydeck_version:
   - '>=0.3'
+python_confluent_kafka_version:
+  - '=1.3.0'
 protobuf_version:
   - '>=3.4.1,<4.0.0'
 pyproj_version:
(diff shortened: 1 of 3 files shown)


conda/recipes/versions.yaml
```diff
@@ -91,6 +91,8 @@ panel_version:
   - '>=0.9.*'
 pydeck_version:
   - '>=0.3'
+python_confluent_kafka_version:
+  - '=1.3.0'
```
kkraus14, line 95: Do we want to hard pin to this version or just say `>=1.3.0`?

jdye64: there are some broker changes in versions above 1.3.0 that would cause breaking changes for our current version of librdkafka. Probably needs to be pinned as we would need to make a few code changes if we used a version above 1.3.0

kkraus14: We're already pinning `librdkafka =1.4.2`

PR #87, opened 2020-07-30
Add gtest, gmock, and benchmark

kkraus14:
Based on https://github.com/rapidsai/cudf/pull/5732

Other libraries will continue to build gtest / gbench from source themselves but opens the doors to copy what cudf is doing

### DIFF
--- a/conda/recipes/versions.yaml
+++ b/conda/recipes/versions.yaml
@@ -26,7 +26,7 @@ build_stack_version:
 arrow_version:
   - '=0.17.1'
 benchmark_version:
-  - '=1.5.0'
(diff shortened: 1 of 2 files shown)


conda/recipes/versions.yaml
```diff
@@ -25,6 +25,8 @@ build_stack_version:
 # Shared versions across meta-pkgs
 arrow_version:
   - '=0.17.1'
+benchmark_version:
+  - '=1.5.0'
```
kkraus14, line 29:
```suggestion
  - '=1.5.1'
```

PR #90, opened 2020-08-05
Add sphinxemoji to meta.yml

galipremsagar:
This will be required to support changes in https://github.com/rapidsai/cudf/pull/5857


raydouglass:
@galipremsagar Looks like `sphinxemoji` doesn't have a conda package. If it is absolutely necessary, we might be able to build a conda package for it.

galipremsagar:
I see, maybe we can hold off for now as I'd be removing `sphinxemoji` for now in PR https://github.com/rapidsai/cudf/pull/5857

Closing this PR for the same reason.

PR #91, opened 2020-08-05
Relax `blas` to allow other BLAS implementations

jakirkham:
We have been pinning to `openblas`. However this causes us issues if we want to install `pytorch`, this runs into issues as they require `mkl`. So this relaxes `blas` to allow either implementation as desired.

cc @mike-wendt @kkraus14 @datametrician

### DIFF
--- a/conda/recipes/rapids-build-env/meta.yaml
+++ b/conda/recipes/rapids-build-env/meta.yaml
@@ -43,7 +43,7 @@ requirements:
     - automake
     - benchmark {{ benchmark_version }}
     - black {{ black_version }}
-    - conda-forge::blas{{ blas_version }}
+    - conda-forge::blas
     - boost
     - boost-cpp {{ boost_cpp_version }}
     - boto3
--- a/conda/recipes/rapids-notebook-env/meta.yaml
+++ b/conda/recipes/rapids-notebook-env/meta.yaml
@@ -37,7 +37,7 @@ requirements:
     - python
   run:
     - bokeh {{ bokeh_version }}
-    - conda-forge::blas{{ blas_version }}
+    - conda-forge::blas
     - cudatoolkit ={{ cuda_version }}.*
     - cupy {{ cupy_version }} # [cuda_compiler_version != "11.0"]
     - cupy {{ cuda11_cupy_version }} # [cuda_compiler_version == "11.0"]
--- a/conda/recipes/versions.yaml
+++ b/conda/recipes/versions.yaml
@@ -37,8 +37,6 @@ benchmark_version:
   - '=1.5.1'
 black_version:
   - '=19.10'
-blas_version:
-  - '=2.14=openblas'
 bokeh_version:
   - '>=2.1.1'
 boost_cpp_version:


kkraus14:
I believe we used to run into problems related to builds where if a numpy build of mkl was pulled it would cause Cython issues, but those issues should be resolved as of now.

I'm not sure if there's other solving issues related to openblas vs mkl though.

jakirkham:
> Seems fine, but are you sure we don't need a minimum blas version?

Yeah it's a good question. Would be good to hear from cuML folks here as well (cc @JohnZed @dantegd @cjnolet).

dantegd:
@jakirkham we required blas because building faiss from source required it, now that we use the conda faiss package we are agnostic AFAIK

datametrician:
> @jakirkham we required openblas explicitly because building faiss from source required it, now that we use the conda faiss package we are agnostic AFAIK

Woot Woot!

mike-wendt:
`mkl` brings in a lot of bulk that is only used by two libraries and they have workarounds. Right now our images create a conda env with `nomkl` and `openblas`to avoid the bloat that `mkl` brings. 

I believe we can merge this and be fine but if the choice is to push for `mkl` then that requires gpuCI image changes as changes here would not solve

mike-wendt:
So with that said is this just relax blas and stay with `openblas` or is there a larger move to support `mkl` despite the downsides and history?

kkraus14:
I think the goal is to just remove the run constraint on allowing a solve to get mkl, i.e. if pytorch is being installed, but we can continue building in gpuCI without mkl to avoid the bloat.

jakirkham:
Yeah exactly as Keith said. I agree it would be better to prefer `openblas` for the reasons noted. Maybe we can force `openblas` in the build case and relax it for other use cases?

mike-wendt:
> I think the goal is to just remove the run constraint on allowing a solve to get mkl, i.e. if pytorch is being installed, but we can continue building in gpuCI without mkl to avoid the bloat.

The solution for anyone in a gpuCI/RAPIDS image to use `pytorch/mkl` is to remove `nomkl` & `openblas` as we do here: https://github.com/rapidsai/cusignal/blob/branch-0.15/ci/gpu/build.sh#L84

I understand that's non-trivial but it is a working path for `cusignal` and `clx` at the moment.

The images are already large to begin with and `mkl` brings in at least another 500MB if not more. I'd prefer if we can stay on `openblas` and maybe make it known more of how to use `mkl` within our images.

Given those constraints what is the consensus on how to move forward?

jakirkham:
Right. We are not forcing `mkl` in this PR. In fact `conda-forge` already treats `openblas` as highest priority when installing `blas` (see below):

```
$ conda create --dry-run -n test_blas blas
Collecting package metadata (current_repodata.json): done
Solving environment: done

## Package Plan ##

  environment location: /datasets/jkirkham/miniconda/envs/test_blas

  added / updated specs:
    - blas


The following NEW packages will be INSTALLED:

  _libgcc_mutex      conda-forge/linux-64::_libgcc_mutex-0.1-conda_forge
  _openmp_mutex      conda-forge/linux-64::_openmp_mutex-4.5-1_llvm
  blas               conda-forge/linux-64::blas-2.17-openblas
  libblas            conda-forge/linux-64::libblas-3.8.0-17_openblas
  libcblas           conda-forge/linux-64::libcblas-3.8.0-17_openblas
  libgcc-ng          conda-forge/linux-64::libgcc-ng-9.3.0-h24d8f2e_14
  libgfortran-ng     conda-forge/linux-64::libgfortran-ng-7.5.0-hdf63c60_14
  liblapack          conda-forge/linux-64::liblapack-3.8.0-17_openblas
  liblapacke         conda-forge/linux-64::liblapacke-3.8.0-17_openblas
  libopenblas        conda-forge/linux-64::libopenblas-0.3.10-pthreads_hb3c22a3_4
  llvm-openmp        conda-forge/linux-64::llvm-openmp-10.0.1-hc9558a2_0



DryRunExit: Dry run. Exiting.
```

All this does is make sure we don't conflict if the user chooses to install `mkl`.

mike-wendt:
@jakirkham good to see this is the case now. My used of `nomkl` is a holdover from years ago when this did not work without me creating a conda env and specifying `nomkl`. I just tested this and we can drop `nomkl` with the change in this PR.

mike-wendt:
> LGTM - just need to file a PR in rapidsai/gpuci-build-environment to remove `nomkl` as well

I'm working on that one and will link here

jakirkham:
Yeah I remember the days of `nomkl`. Things have definitely improved since. Thanks Mike! 😄

PR #93, opened 2020-08-06
Require Cython 0.29.14+

jakirkham:
This is needed as `PyMemoryView_GET_BUFFER` was added to the `cpython.memoryview` API in Cython's standard library, which is used by `ucx-py`.

xref: https://github.com/rapidsai/ucx-py/pull/384
xref: https://github.com/cython/cython/commit/6b01c3725c4a6b488332d76a305b2d349953d0b9

### DIFF
--- a/conda/recipes/versions.yaml
+++ b/conda/recipes/versions.yaml
@@ -52,7 +52,7 @@ cmake_setuptools_version:
 cupy_version:
   - '>=7.1.0,<8.0.0a0'
 cython_version:
-  - '>=0.29,<0.30'
+  - '>=0.29.14,<0.30'
 dask_version:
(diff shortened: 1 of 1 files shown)


jakirkham:
cc @mike-wendt

jakirkham:
rerun tests

jakirkham:
Failure is due to CUDA 11. Not sure if that is expected.

mike-wendt:
> Failure is due to CUDA 11. Not sure if that is expected.

This is expected on `meta-pkgs` until we have all of the CUDA 11 pkgs needed by the `rapids` pkg. We only have some so this is broken for the time being.

PR #94, opened 2020-08-06
[WIP] Drop ucx-py

jakirkham:
Drops `ucx-py` to avoid having the package show up in `ucx-py`'s CI. Also explicitly lists `ucx`, which we need in all cases ( including `ucx-py` going forward https://github.com/rapidsai/ucx-py/pull/570 ).

Note: Marked as draft until we fix downstream projects to include this explicitly.

cc @mike-wendt

### DIFF
--- a/conda/recipes/rapids-build-env/meta.yaml
+++ b/conda/recipes/rapids-build-env/meta.yaml
@@ -132,7 +132,6 @@ requirements:
     - twine
     - ucx-proc=*=gpu
     - ucx
-    - ucx-py ={{ minor_version }}.*
     - umap-learn
 
 about:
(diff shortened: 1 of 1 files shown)


jakirkham:
[Searching RAPIDS]( https://github.com/search?l=Python&p=2&q=org%3Arapidsai+ucp+-repo%3Arapidsai%2Fucx-py&type=Code ) it looks like `cudf`, `cuml`, `dask-cuda`, and `raft` will need this. Though please let me know if you spot others.

pentschev:
I'm +1 on this, thanks @jakirkham .

jakirkham:
Looks like the failure here is only on CUDA 11. IDK if that is expected.

mike-wendt:
> Looks like the failure here is only on CUDA 11. IDK if that is expected.

As mentioned in #93 this is expected for now.

PR #95, opened 2020-08-07
FIX Add 0.16 RAPIDS to nightly axis & remove `libcumlprims` from `rapids-build-env`

mike-wendt:
This enables env/meta-pkg builds for v0.16 nightlies as burndown started yesterday.

It also removes `libcumlprims` from `rapids-build-env` as `libcumlprims` needs a `0.16` container to build which is blocked by getting these env-pkgs out. `cuML` already explicitly installs `libcumlprims` so removing this package should have no impact.

Blocks rapidsai/gpuci-build-environment#116
Blocks rapidsai/docker#138 (which is blocked by the above PR)

### DIFF
--- a/ci/axis/nightly.yaml
+++ b/ci/axis/nightly.yaml
@@ -6,6 +6,7 @@ CONDA_CONFIG_FILE:
 
 RAPIDS_VER:
   - 0.15
+  - 0.16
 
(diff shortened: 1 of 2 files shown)


mike-wendt:
rerun tests

PR #96, opened 2020-08-07
FIX Correct conda recipes to use RAPIDS version

mike-wendt:
Currently the recipes use `version` which is based off of the tag in this repo. Switching this to `rapids_version` which uses the env var `RAPIDS_VER` specified by `./ci/axis/nightly.yaml` or `./ci/axis/release.yaml`. This way during burndown where we need to build multiple version from the `nightly.yaml` axis they will each get their respective version.

As it is right now they get the same version so while a `0.16` build shows as successful it is versioned as `0.15` and has a duplicate package.

PR #97, opened 2020-08-07
FIX Update RAPIDS versions in axis files

mike-wendt:
With #96 RAPIDS versions need to be full versions like `0.15.0a` instead of `0.15`. This is because we are no longer using the git tag to pull this information and without it we get conda files with the version `0.15` instead of matching the nightly version of `0.15.0a`.

Additionally cleaned up our use of git tags to set version as the axis files setting `RAPIDS_VER` will control the versioning in this repo.

### DIFF
--- a/ci/axis/nightly.yaml
+++ b/ci/axis/nightly.yaml
@@ -5,8 +5,8 @@ CONDA_CONFIG_FILE:
   - conda/recipes/versions.yaml
 
 RAPIDS_VER:
-  - 0.15
-  - 0.16
+  - 0.15.0a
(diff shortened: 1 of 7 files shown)


mike-wendt:
Verified new versioning works `rapids-build-env-0.15.0a-cuda11.0_py37_gff181f0_174.tar.bz2`

mike-wendt:
Will merge when tests complete and spot checks are done. Again `gpuCI/build/meta-pkg-nightly` is an expected fail for `0.16` and CUDA 11

PR #98, opened 2020-08-07
FIX Include VERSION_SUFFIX in pkg version

mike-wendt:
PR tests showed the correct version; however, they were missing the `VERSION_SUFFIX` which for branch builds is added in the form `YYMMDD` to produce `0.15a200807`. This update fixes the versioning to match and patches #97 and #96.


mike-wendt:
Will merge when tests complete and spot checks are done. Again `gpuCI/build/meta-pkg-nightly` is an expected fail for `0.16` and CUDA 11

PR #105, opened 2020-08-12
FIX Update test PATHs and remove duplicate installs

mike-wendt:
Jenkins on docker runs overwrites the container's `PATH` with the node's `PATH`. These patches reinstates the correct `PATH` prior to `conda activate`.

In addition, there are a number of conda/pip installs that are no longer necessary and are removed.


mike-wendt:
Ran an rmm test with this and it was successful for 5/6 - https://gpuci.gpuopenanalytics.com/job/rapidsai/job/nightly-tests/job/docker-test-rmm/718/

> One centos7 test is missing a CSV file

The others are 2-4 hrs long with 6 jobs each so I don't want to clog the pipes at the moment with their tests

PR #110, opened 2020-08-17
FIX cuDF Test Wrong GCC

aleksficek:
cuDF builds not working on centos7 after recent nvcc PATH fix

The `PATH` variable needed to be updated to include `the gcc7` location when on centos7.

https://gpuci.gpuopenanalytics.com/job/rapidsai/job/nightly-tests/job/docker-test-cudf/

Open to ideas if there's a cleaner way to do this, this path works fine for ubuntu because `usr/local/gcc7` does not exist on it and is therefore skipped.


aleksficek:
Closing as this issue should be fixed in the docker repo

aleksficek:
Reopening as this might actually be an integration repo solution.

PR #109, opened 2020-08-17
ENH Update numba versioning to pull 0.51

mike-wendt:
From successful GPU unit tests in rapidsai/cudf#5979 we are going to remove the restriciton that prevents use of numba 0.51 given the earlier timeout issues are resolve. Numba 0.51 brings CUDA 11 and Ampere support.

PR #112, opened 2020-08-18
datashader version supporting cuDF 0.14 and above

AjayThorve:
Updating datashader version to 0.11.1 to resolve issues supporting cuDF 0.14 and above. This also requires numba >= 0.51.2, dependent on #109 being resolved.


mike-wendt:
#109 was just merged - that said we need to wait for a `0.51.1` package to be published to merge this

Still waiting to get an idea on timing for this and if there will be a pkg shortly that we can use - stay tuned

mike-wendt:
rerun tests

PR #99, opened 2020-08-08
Upgrade cmake to 3.17+

kkraus14:
cc @aleksficek as I'm guessing this has ccache integration interactions

CMake 3.17+ is needed to use new `FindCUDAToolkit` features which are required for https://github.com/rapidsai/rmm/pull/474 and will be required for all RAPIDS packages in the future.

We want to target 0.16 for this as opposed to 0.15.

PR #117, opened 2020-08-25
[REVIEW] bump cmake to 3.17

kkraus14:
See #99

### DIFF
--- a/conda/recipes/versions.yaml
+++ b/conda/recipes/versions.yaml
@@ -44,7 +44,7 @@ boost_cpp_version:
 clang_version:
   - '=8.0.1'
 cmake_version:
-  - '=3.14.5'
+  - '=3.18.0'
 cmake_setuptools_version:
   - '>=0.1.3'
 cupy_version:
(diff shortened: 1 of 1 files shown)


mike-wendt:
@kkraus14 are you ready for this in `0.16` nightlies today?

kkraus14:
> @kkraus14 are you ready for this in `0.16` nightlies post-release?

There will be a ton of spurious warnings that we'll need to address across all of the libs, but there may be a way for us to ignore the warning. Let me look into it.

cc @JohnZed @BradReesWork to confirm this should be okay from the cuML / cuGraph side.

mike-wendt:
@BradReesWork can you approve?

kkraus14:
@mike-wendt I think we need to hold off a bit on this and we should only bump to 3.17 to start. @galipremsagar and @rgsl888prabhu have reported that with 3.18 it causes it to always rebuild the entire libcudf library, so we need to investigate.

kkraus14:
Updated PR to use CMake 3.17 instead.

PR #121, opened 2020-08-28
pinning doxygen version to avoid potential confusion

teju85:
This helps avoid surprises like the one that happened with cuML here: https://github.com/rapidsai/cuml/pull/2769.

### DIFF
--- a/conda/recipes/rapids-doc-env/meta.yaml
+++ b/conda/recipes/rapids-doc-env/meta.yaml
@@ -35,7 +35,7 @@ requirements:
     - pip
   run:
     - beautifulsoup4
-    - doxygen
+    - doxygen=1.8.20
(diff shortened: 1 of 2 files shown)


conda/recipes/rapids-doc-env/meta.yaml
```diff
@@ -35,7 +35,7 @@ requirements:
     - pip
   run:
     - beautifulsoup4
-    - doxygen
+    - doxygen=1.8.20
```
dillon-cullinan, line 38: Version control is determined with the `versions.yml` file under `conda/recipes`.

teju85: Ah.. missed that one. Thanks. Fixed it.

PR #130, opened 2020-09-15
[REVIEW]Add nltk to rapids-build-env

VibhuJawa:
Add nltk to rapids-build-env/meta.yaml . 

We need `nltk` for adding a test for the porter stemmer implementation in PR https://github.com/rapidsai/cuml/pull/2476 .


VibhuJawa:
Unsure why it failed for `0.15` but i don't think it should be related to the PR. 


image

ajschmidt8 (approved):

ajschmidt8:
> Unsure why it failed for `0.15` but i don't think it should be related to the PR.

That can be ignored (fixed in #131).

PR #143, opened 2020-10-06
Remove dask-xgboost

raydouglass:
dask-xgboost was deprecated in v0.15 and to be removed in v0.16.

PR #145, opened 2020-10-07
[REVIEW] pin faiss to =1.6.3

dantegd:
faiss had some file re-organization, so there is a good chance that the next release could be incompatible with cuML as it is right now.


kkraus14 (approved):

mike-wendt:
Will retry testing once #143 is merged with a fix for the 0.17 failures

mike-wendt:
rerun tests

mike-wendt (approved):

PR #150, opened 2020-10-09
Use `matplotlib-base` instead of `matplotlib`

jakirkham:
The `matplotlib` package ships with a `qt` dependency, which we don't need or use in the notebook. Not too mention this is a quite heavyweight dependency. So use `matplotlib-base` instead. This has just enough of the goodies to allow matplotlib to work from a notebook or usable as a dependency. Should help lighten our containers and simplify installs for people.


kkraus14 (approved):

ajschmidt8 (approved):

jakirkham:
Thanks all! 😄

PR #170, opened 2020-11-06
[REVIEW] updated clang version to 11.0.0

teju85:
## Why update?
If we want to enable running clang-tidy on .cu files, v8.0.1 does not even support CTK version 10, as described here: https://github.com/rapidsai/raft/pull/85. Thus, we need to update clang version.

## Any risks due to this update?
None so far. I have tested this by running clang-tidy and clang-format on both cuML and raft repo's:
1. https://github.com/rapidsai/cuml/pull/3121
2. https://github.com/rapidsai/raft/pull/89

Specifically, the set of changes caused by clang-format is almost negligible.

### DIFF
--- a/conda/recipes/versions.yaml
+++ b/conda/recipes/versions.yaml
@@ -34,7 +34,7 @@ bokeh_version:
 boost_cpp_version:
   - '=1.72.0'
 clang_version:
-  - '=8.0.1'
+  - '=11.0.0'
 cmake_version:
   - '=3.17'
 cmake_format_version:


teju85:
@codereport, can you check if the above claims are true for cudf too? (meaning, clang update doesn't cause much of clang-format violations)

teju85:
@codereport also tried upgrading to clang-11 in his PR here: https://github.com/rapidsai/cudf/pull/6695 and saw not much changes in cuDF too. So, it seems like it is safer to upgrade clang version on our meta recipe.

ajschmidt8 (approved):

teju85:
Tagging relevant folks to let us know if they are ok with this version upgrade:
1. for cugraph, @afender or @seunghwak 
2. for rmm, cuspatial @harrism or @codereport

mike-wendt:
From @kkraus14 punt to 0.18

raydouglass:
@teju85 Is this still needed?

teju85:
@raydouglass Yes. This is still needed. However, we require a "go ahead" signal from the above code owners to proceed.

ajschmidt8:
superseded by #390. closing

PR #188, opened 2020-12-09
Enable parallelism in cudf & dask-cudf pytests

galipremsagar:
This PR enables pytest-xdist in cudf and dask-cudf pytests.

### DIFF
--- a/ci/test/cudf.sh
+++ b/ci/test/cudf.sh
@@ -50,7 +50,7 @@ if (( ${exitcode} != 0 )); then
 fi
 
 cd /rapids/cudf/python/custreamz
-py.test --junitxml=${TESTRESULTS_DIR}/pytest-custreamz.xml -v
+py.test -n 6 --junitxml=${TESTRESULTS_DIR}/pytest-custreamz.xml -v
 exitcode=$?
 if (( ${exitcode} != 0 )); then
    SUITEERROR=${exitcode}
--- a/ci/test/cudf.sh
+++ b/ci/test/cudf.sh
@@ -34,7 +34,7 @@ export PYTHONPATH=\
 ${PYTHONPATH}
 
 cd /rapids/cudf/python/cudf
-py.test --junitxml=${TESTRESULTS_DIR}/pytest-cudf.xml -v
+py.test -n 6 --junitxml=${TESTRESULTS_DIR}/pytest-cudf.xml -v
 exitcode=$?
 if (( ${exitcode} != 0 )); then
    SUITEERROR=${exitcode}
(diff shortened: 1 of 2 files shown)


ci/test/cudf.sh
```diff
@@ -34,15 +34,15 @@ export PYTHONPATH=\
 ${PYTHONPATH}
 
 cd /rapids/cudf/python/cudf
-py.test --junitxml=${TESTRESULTS_DIR}/pytest-cudf.xml -v
+py.test -n 6 --junitxml=${TESTRESULTS_DIR}/pytest-cudf.xml -v
```
galipremsagar, line 37: Since the CI jobs have 8vCPUs, I've left out 2 cores for pytest to not choke.

galipremsagar: cc: @raydouglass @mike-wendt

galipremsagar:
I've locally tested pytests with enabling cuda MPS, there isn't any noticeable speeds just by using MPS but however it would help us with CUDA context switching when there are multiple processes. 

@rapidsai/ops-codeowners Could you point where the following set of commands need to be added in this repo? These set of commands except running pytests need to be executed as a root user.

# Execute as root

export CUDA_VISIBLE_DEVICES="0"
nvidia-smi -i 0 -c EXCLUSIVE_PROCESS
nvidia-cuda-mps-control -d


# Run Pytests
...
...

# MPS Cleanup

echo quit | nvidia-cuda-mps-control
nvidia-smi -i 0 -c DEFAULT

kkraus14:
Can we handle MPS in a follow up? Is there a net benefit without MPS?

galipremsagar:
> Can we handle MPS in a follow up? Is there a net benefit without MPS?

Sure this can be done as a follow-up and maybe we can add MPS if we see context issues but currently I did not notice any difference in execution times. I've been using parallel pytests without enabling MPS for quite a very long time. So parallel pytests should run properly without enabling MPS too.

raydouglass:
I don't think we can enable MPS as running commands as root is not possible inside of the gpuCI docker containers.

galipremsagar:
> running commands as root is not possible inside of the gpuCI docker containers.

Didn't know that. In that case, we probably go ahead without MPS too.

mike-wendt:
> I don't think we can enable MPS as running commands as root is not possible inside of the gpuCI docker containers.

Due to the security concerns this is not an option in gpuCI. If there is a way to do this without `root` access we can make it work; otherwise, we can only run things are a privileged user to limit system access as these run on a shared environment

kkraus14:
IIRC this is a driver level setting that interacts with the kernel so you can't really run it within a container even with root unless you use the `--privileged` flag which is an obvious no-no.

Lets move forward without MPS for now and can explore it in the future.

mike-wendt:
rerun tests

mike-wendt:
Blocked until `runtime` images are fixed

mike-wendt:
rerun tests

mike-wendt:
rerun tests

mike-wendt (approved):

PR #190, opened 2020-12-09
FIX Add rapids-pytest-benchmark pkg for cugraph tests

mike-wendt:
This fixes failing nightly tests for cugraph due to:
```
 + py.test --junitxml=/var/lib/jenkins/workspace/rapidsai/nightly-tests/docker-test-cugraph/CUDA_VER/11.0/DOCKER_REPO/rapidsai/rapidsai-dev-nightly/LINUX_VER/ubuntu16.04/PYTHON_VER/3.7/RAPIDS_VER/0.17/testresults/pytest.xml -v /rapids/cugraph/python
 ERROR: usage: py.test [options] [file_or_dir] [file_or_dir] [...]

 py.test: error: unrecognized arguments: --benchmark-gpu-disable
   inifile: /rapids/cugraph/python/pytest.ini
   rootdir: /rapids/cugraph/python
```

PR #191, opened 2020-12-09
FIX Add rapids-pytest-benchmark pkg for cugraph tests

mike-wendt:
Merging changes into branch-0.18 from #190


ajschmidt8 (approved):

mike-wendt:
Will open a new PR shortly that does not impact the previous

PR #193, opened 2020-12-09
ENH Remove CUDA default conda builds

mike-wendt:
These CUDA default pkgs are no longer used as we specify the CUDA versions with the install. They currently result in a longer build time with no benefit. Removing these extra builds to reduce build time


mike-wendt:
Will need a follow up PR to remove the `DEFAULT_CUDA_VER` from the axis files; doing so now would break the current PRs in progress - will create an issue for this and add notes to the axis files for now so we can reclaim the unnecessary build time

PR #201, opened 2020-12-30
Upgrade to Treelite 1.0.0rc1

hcho3:
Required by rapidsai/cuml#3316, which needs the 1.0.0 version of the Treelite package.

For now the RC1 version is used; once the testing is complete with rapidsai/cuml#3316, I will release the 1.0.0 version.

### DIFF
--- a/conda/recipes/versions.yaml
+++ b/conda/recipes/versions.yaml
@@ -137,4 +137,4 @@ spdlog_version:
 sphinx_markdown_tables_version:
   - '=0.0.14=pyh9f0ad1d_1'
 treelite_version:
-  - '=0.93'
+  - '=1.0.0rc1'
--- a/conda/recipes/versions.yaml
+++ b/conda/recipes/versions.yaml
@@ -21,7 +21,7 @@ conda_verify_version:
 
 # Versions for `libgcc-ng`, `libgfortran-ng`, `libstdcxx-ng` stack
 build_stack_version:
-  - '=7.5.0'
+  - '=9.3.0'
 
 # Shared versions across meta-pkgs
 arrow_version:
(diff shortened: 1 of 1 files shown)


ajschmidt8 (dismissed):

ajschmidt8:
image

I think the updated `libgcc-ng` version that's used in the new `treelite` package (shown above) may be conflicting with the [libgcc-ng version that we use](https://github.com/rapidsai/integration/blob/branch-0.18/conda/recipes/versions.yaml#L22-L24). I'm not sure the best way to resolve this. I'll have to defer to @mike-wendt or @raydouglass when they're back from holiday next week since they're more familiar with our stack.

hcho3:
@ajschmidt8 Conda-forge upgraded their compiler stack to GCC 9 back in October: https://conda-forge.org/docs/user/announcements.html

cc @jakirkham

jakirkham:
I think PR ( https://github.com/conda-forge/treelite-feedstock/pull/20 ) should let us use the older compilers for now. Though we might want to start planning to move to GCC 9 where possible in RAPIDS in the future.

conda/recipes/versions.yaml
```diff
@@ -21,7 +21,7 @@ conda_verify_version:
 
 # Versions for `libgcc-ng`, `libgfortran-ng`, `libstdcxx-ng` stack
 build_stack_version:
-  - '=7.5.0'
+  - '=9.3.0'
```
jakirkham, line 24: Think we can revert this now with the new `treelike` packages built with the old compilers.

```suggestion
  - '=7.5.0'
```

hcho3: Reverted

ajschmidt8 (approved):

PR #209, opened 2021-01-21
update panel version dependency

AjayThorve:
Updated panel dependency to match recent cuxfilter update(https://github.com/rapidsai/cuxfilter/pull/226) to `>=0.10.3` 

cc @ajschmidt8

### DIFF
--- a/conda/recipes/versions.yaml
+++ b/conda/recipes/versions.yaml
@@ -106,7 +106,7 @@ pandas_version:
 pandoc_version:
   - '<=2.0.0'
 panel_version:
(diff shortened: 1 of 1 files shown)


dillon-cullinan (approved):
Are there any other repos that depend on panel other than cuSpatial?

ajschmidt8 (approved):

ajschmidt8:
> Are there any other repos that depend on panel other than cuXfilter?

Doesn't look like it.
src: https://github.com/search?l=YAML&q=org%3Arapidsai+panel&type=Code

PR #239, opened 2021-04-05
Remove `rapids-pytest-benchmark` dependency

raydouglass:
Since `rapids-pytest-benchmark` now depends on `rmm`, this pollutes the `gpuci/rapidsai` images with an install of RMM on the same image we need to build RMM.

Jobs/repos that require this package for benchmarks will need to manually install install it instead of relying on `rapids-build-env` to incude it.


dillon-cullinan (approved):

ajschmidt8 (approved):

raydouglass:
Meta packages are expected failure. Admin merging...

PR #245, opened 2021-04-08
Add ccache to the rapids-build-env meta package

teju85:
Since I didn't find a `branch-0.20` where I could send a PR for this, I'm using 0.19 instead.

### DIFF
--- a/conda/recipes/rapids-build-env/meta.yaml
+++ b/conda/recipes/rapids-build-env/meta.yaml
@@ -47,6 +47,7 @@ requirements:
     - boost-cpp {{ boost_cpp_version }}
     - boto3
     - cachetools
+    - ccache {{ ccache_version }}
     - conda-forge::clang {{ clang_version }}
     - conda-forge::clang-tools {{ clang_version }}
     - cmake {{ cmake_version }}
--- a/conda/recipes/versions.yaml
+++ b/conda/recipes/versions.yaml
@@ -34,6 +34,8 @@ bokeh_version:
   - '>=2.1.1,<=2.2.3'
 boost_cpp_version:
   - '=1.72.0'
+ccache_version:
+  - '=4.1'
 clang_version:
   - '=8.0.1'
 cmake_version:


raydouglass:
@dillon-cullinan Can you review this?

dillon-cullinan (commented):
Is there more context for the necessity of this? We currently use our own custom release package named `gpuci-ccache` which is currently using the latest stable release of `ccache`. I would also argue this does not belong in the build environment as `ccache` is not required to build the RAPIDS suite.

If you want the compatible `ccache` that we use in CI, you can utilize our package

```
conda install -c gpuci -c conda-forge gpuci-ccache
```

teju85:
I wasn't aware of `gpuci-ccache`! @dillon-cullinan would you also be opposed to putting this instead of `ccache` in our build-env?

My intentions with adding this to build-env are:
1. As noticed by many others in RAPIDS team, `ccache` has the ability to significantly reduce build times (especially for folks who do `make clean && make` frequently). Eg: in the case of cuML, with this enabled, a clean build only took ~12s on a DGX1V cpu!
2. cuML's [`build.sh`](https://github.com/rapidsai/cuml/blob/branch-0.19/build.sh) already supports building with `ccache`. So, I'm hopeful of dev's from other teams also starting to use this feature.

> I would also argue this does not belong in the build environment as ccache is not required to build the RAPIDS suite.

Not sure I quite follow your argument. Sure it is *currently* not required. But we all know its benefits and it is just a matter of time that we all decide to use it, right? (Yes, maybe not directly `ccache`, but atleast `gpuci-ccache`). Or am I missing something here?

Let me tag @JohnZed and @kkraus14 also to get their opinions on using `ccache` in our dev's build-env in all of RAPIDS C/C++ libs.

teju85:
Never mind guys. Many of the things got clarified in an offline thread with @rlratzel and @dantegd . Since this meta pkg is mostly for CI env, and we currently do not support ccache based builds in our CI. I now get @dillon-cullinan 's points above.

I'm ok with just closing this PR, unless anybody else has things to say otherwise.

ajschmidt8:
We're currently in the process of switching our CI system to use `sccache` instead of `ccache`, so I think it's safe to close this issue.

PR #264, opened 2021-04-29
Bump CMake version to 3.20.1

wphicks:
Bring in CMake version which will facilitate transition to rapids-cmake and eliminate a variety of bugs related to CUDA, ccache, and compiler detection

### DIFF
--- a/conda/recipes/versions.yaml
+++ b/conda/recipes/versions.yaml
@@ -37,7 +37,7 @@ boost_cpp_version:
 clang_version:
   - '=8.0.1'
 cmake_version:
-  - '=3.20.1'
+  - '>=3.20.1,<3.21'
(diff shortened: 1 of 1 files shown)


conda/recipes/versions.yaml
```diff
@@ -37,7 +37,7 @@ boost_cpp_version:
 clang_version:
   - '=8.0.1'
 cmake_version:
-  - '=3.18'
+  - '=3.20.1'
```
jakirkham, line 40: Would suggest letting the patch version float (though we can set a lower limit if that's helpful)

```suggestion
  - '>=3.20.1,<3.21'
```

raydouglass:
rerun tests

ajschmidt8 (approved):

PR #265, opened 2021-05-05
Upgrade to Treelite 1.2.0

hcho3:
Required by rapidsai/cuml#3829. See the CI result in rapidsai/cuml#3829 which is testing cuML with Treelite 1.2.0.


hcho3:
rapidsai/cuml#3829 revealed a critical bug in Treelite, which will be address in the upcoming patch release of Treelite (1.2.1). I will re-submit once 1.2.1 is released.

PR #266, opened 2021-05-05
Add JupyterLab 3 Dependencies

ajschmidt8:
This PR adds a few dependencies to our packages following the upgrade to JupyterLab 3 in #261 and https://github.com/rapidsai/jupyterlab-nvdashboard/pull/85.

The `python-build` package is now needed to build the `nvdashboard` extension.

Additionally, the dependencies listed [here](https://github.com/rapidsai/docker/blob/f23d1f18dd79639eaad3dc580fa946997f8e1524/templates/rapidsai-core/partials/install_notebooks.dockerfile.j2#L10-L15) can now be installed solely via `conda` and no longer need the `jupyter labextension install` counterpart.

### DIFF
--- a/conda/recipes/rapids-build-env/meta.yaml
+++ b/conda/recipes/rapids-build-env/meta.yaml
@@ -125,7 +125,6 @@ requirements:
     - pytest-timeout
     - pytest-xdist
     - python
-    - python-build
     - python-confluent-kafka {{ python_confluent_kafka_version }}
     - python-louvain
     - rapidjson {{ rapidjson_version }}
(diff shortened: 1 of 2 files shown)


conda/recipes/rapids-build-env/meta.yaml
```diff
@@ -125,6 +125,7 @@ requirements:
     - pytest-timeout
     - pytest-xdist
     - python
+    - python-build
```
dillon-cullinan, line 128: This should be moved to the notebook environment in my opinion. We don't need this to build RAPIDS, and it makes sense in the notebook env since its required for nvdashboard.

dillon-cullinan (changes requested):

conda/recipes/rapids-build-env/meta.yaml
ajschmidt8: done

dillon-cullinan (approved):

PR #268, opened 2021-05-06
Remove python-build dependency

ajschmidt8:
This PR removes the `python-build` dependency that was added in #266. It is no longer needed for the `jupyterlab-nvdashboard` repo after switching from using `python -m build` to `python setup.py sdist bdist_wheel` to build and package the extension in https://github.com/rapidsai/jupyterlab-nvdashboard/pull/95.


ajschmidt8:
Keeping this as a draft until https://github.com/rapidsai/jupyterlab-nvdashboard/pull/95 is merged since it's still pending approvals.

dillon-cullinan (approved):

PR #271, opened 2021-05-14
Bump geopandas version to match cuspatial requirements.

thomcom:
Geopandas is 0.9.0+ in cuspatial, now.

### DIFF
--- a/conda/recipes/versions.yaml
+++ b/conda/recipes/versions.yaml
@@ -71,7 +71,7 @@ fsspec_version:
 gdal_version:
   - '>=3.2.0,<3.3.0a0'
 geopandas_version:
-  - '>=0.9.0'
+  - '>=0.9.0,<0.10.0a0'
 gcsfs_version:
   - '>=0.8.0'
 google_cloud_cpp_version:
--- a/conda/recipes/versions.yaml
+++ b/conda/recipes/versions.yaml
@@ -71,7 +71,7 @@ fsspec_version:
 gdal_version:
   - '>=3.2.0,<3.3.0a0'
 geopandas_version:
-  - '>=0.6, <=0.8.1'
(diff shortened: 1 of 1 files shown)


conda/recipes/versions.yaml
```diff
@@ -71,7 +71,7 @@ fsspec_version:
 gdal_version:
   - '>=3.2.0,<3.3.0a0'
 geopandas_version:
-  - '>=0.6, <=0.8.1'
+  - '>=0.9.0'
```
kkraus14, line 74: I believe in cuspatial we pinned it to `>=0.9.0,<0.10.0a0` since geopandas is not yet API stable.

thomcom: Thanks for pointing this out. In `cuspatial/meta.yaml` we're using as you specified, in `environments/cuspatial_dev_cuda*.yml` we're using what I specified. This is a discrepancy I've been seeing lately - shouldn't the conda dev environments use the same recipe as the conda release environment from `meta.yaml`? I'd like to get rid of the `.yml` files if possible.

thomcom: Also, fixed as you requested.

kkraus14 (approved):

thomcom:
I don't see why the checks are not successful. Console logs don't reveal anything. @ajschmidt8 ?

ajschmidt8:
looks like the failures are related to the calver transition. specifically, we were still waiting on some Docker images and the xgboost package. the xgboost package is available now and the Docker images are currently building. once the Docker images are published, I'll rerun CI here and hopefully we'll get some passing checks!

thomcom:
rerun tests

ajschmidt8:
rerun tests

ajschmidt8 (approved):

PR #288, opened 2021-06-01
[REVIEW] Adding hdbscan for cuml test dependency

cjnolet:


dantegd (approved):

cjnolet:
Looks like this is unhappy about an XGboost package:
```
11:33:55   - xgboost=1.4.0dev.rapidsai21.08
```

ajschmidt8 (approved):

PR #291, opened 2021-06-04
Add `nbsphinx` version spec

ajschmidt8:
This PR adds a version specifier to the `nbsphinx` package. The version should be a minimum of `0.8.6` to fix the doc build issues described in https://github.com/rapidsai/cudf/issues/8345.

### DIFF
--- a/conda/recipes/versions.yaml
+++ b/conda/recipes/versions.yaml
@@ -96,6 +96,8 @@ mypy_version:
   - '0.782'
 mysql_connector_cpp_version:
   - '8.0.23'
+nbsphinx_version:
+  - '>=0.8.6'
 nccl_version:
   - '>=2.9.9,<3.0a0'
 networkx_version:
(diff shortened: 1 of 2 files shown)


dillon-cullinan (approved):

conda/recipes/versions.yaml
```diff
@@ -96,6 +96,8 @@ mypy_version:
   - '0.782'
 mysql_connector_cpp_version:
   - '8.0.23'
+nbsphinx_version:
+  - '>=0.8.6'
```
jakirkham, line 100: Should we add a max version as well?

jakirkham (commented):
Looks good. Thanks AJ! 😀

Had one question

conda/recipes/versions.yaml
ajschmidt8: If the patch version bump from `0.8.5` to `0.8.6` broke our builds, I'm not sure what would be a safe maximum version here :stuck_out_tongue:. Pinning to only `0.8.6` seems too restrictive. Probably fine to omit it for this package. We can add a maximum if it becomes a problem again.

jakirkham: Sounds good :)

PR #292, opened 2021-06-07
Ignore `raft` unit tests for `cugraph` [skip ci]

ajschmidt8:
This PR adds a flag to ignore the `raft` unit tests in the `cugraph` repo since those tests are meant to be run from the `raft` repo itself (and not a downstream repo).

### DIFF
--- a/ci/test/cugraph.sh
+++ b/ci/test/cugraph.sh
@@ -34,7 +34,7 @@ for gt in /rapids/cugraph/cpp/build/tests/*_TEST; do
 done
 
 # Python tests
-py.test --junitxml=${TESTRESULTS_DIR}/pytest.xml -v /rapids/cugraph/python
+pytest --ignore=cugraph/raft --junitxml=${TESTRESULTS_DIR}/pytest.xml -v /rapids/cugraph/python
 exitcode=$?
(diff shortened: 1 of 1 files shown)


rlratzel (approved):
LGTM. It would be nice if we could have a universal test script for cugraph (which was the intent behind [this](https://github.com/rapidsai/cugraph/blob/branch-21.08/ci/test.sh)), much like we do with `build.sh` but that might be out-of-scope for this PR - just something to keep in mind to reduce maintenance later.

dillon-cullinan (approved):

dillon-cullinan:
> LGTM. It would be nice if we could have a universal test script for cugraph (which was the intent behind [this](https://github.com/rapidsai/cugraph/blob/branch-21.08/ci/test.sh)), much like we do with `build.sh` but that might be out-of-scope for this PR - just something to keep in mind to reduce maintenance later.

We have the source builds of every repo in the images we are testing with this, so its very doable. Though I would like it to be fleshed out with all the repos.

That way we can either have a generalized script in here that references the repos via a variable and we just have a common structure and execution, which also results in us being able to combine all the nightly devel jobs into a single job.

PR #297, opened 2021-06-09
Fix `cugraph` flag path

ajschmidt8:
This flag path from #292 needs to be absolute, not relative.


dillon-cullinan (approved):

ajschmidt8:
admin merging since this change isn't tested in the GitHub checks.

PR #303, opened 2021-06-24
[REVIEW] Upgrade arrow & librdkafka

galipremsagar:
This PR bumps arrow version to `4.0.1` and we would also need `librdkafka` to be fetching latest versions of `1.6.x` as older versions bring in a dependency `lz4-c` that would be incompatible. 

Needed for : https://github.com/rapidsai/cudf/pull/7495/

### DIFF
--- a/conda/recipes/versions.yaml
+++ b/conda/recipes/versions.yaml
@@ -25,7 +25,7 @@ build_stack_version:
 
 # Shared versions across meta-pkgs
 arrow_version:
-  - '=1.0.1'
+  - '=4.0.1'
 benchmark_version:
   - '=1.5.1'
 black_version:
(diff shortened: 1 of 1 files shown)


conda/recipes/versions.yaml
```diff
@@ -89,7 +89,7 @@ jupyterlab_version:
 jupyter_packaging_version:
   - '>=0.7.0,<0.8'
 librdkafka_version:
-  - '=1.5.*'
+  - '=1.6.*'
```
jdye64, line 92: LFTM. We only use core librdkafka features that have not changed in several releases so a bump like this is no issue.

jdye64 (approved):

raydouglass (approved):

kkraus14 (approved):

PR #317, opened 2021-07-20
[REVIEW] ENH Add 21.10 meta packages to nightly builds

dillon-cullinan:
Adds 21.10 to the axis to get the next version of meta packages published. This is required for 21.10 images to be released, which is blocking newly created PRs to the latest dev branch.

### DIFF
--- a/ci/axis/nightly.yaml
+++ b/ci/axis/nightly.yaml
@@ -7,6 +7,7 @@ CONDA_CONFIG_FILE:
 # Use M.X.Ya (major.minor.patch) with 'a' version to match nightly tag
 RAPIDS_VER:
   - 21.08.00a
+  - 21.10.00a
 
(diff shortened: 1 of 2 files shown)


ajschmidt8 (dismissed):

raydouglass (changes requested):
This isn't necessary since `branch-21.10` exists with these changes

dillon-cullinan:
> This isn't necessary since `branch-21.10` exists with these changes

Our Jenkins jobs in CI don't build on this branch, though, and those jobs are all manually written, so this changes the process quite a bit.

If we don't use the axis file to handle this, we need to create another Jenkins job that handles the `ver+1` pipeline. Then enable/disable that job as needed between release cycles. I'm not really sure if this is a better approach than just updating axis files. I suppose ideally we would have generated jobs like gpuci-scripts, but that isn't really in place right now.

PR #318, opened 2021-07-21
Pinning spdlog version to 1.8.5

cjnolet:
A new conda package for spdlog was released last night and it's causing a build issue in cuml. We've decided to pin the version to 1.8.5 in the meantime to unblock CI (and potentially for release 21.08). 

https://github.com/rapidsai/cuml/issues/4078

### DIFF
--- a/conda/recipes/versions.yaml
+++ b/conda/recipes/versions.yaml
@@ -149,7 +149,7 @@ scipy_version:
 setuptools_version:
   - '>=49,<50'
 spdlog_version:
-  - '>=1.8.5,<2.0.0a0'
+  - '=1.8.5'
 sphinx_markdown_tables_version:
   - '=0.0.14=pyh9f0ad1d_1'
 treelite_version:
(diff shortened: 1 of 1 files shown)


conda/recipes/versions.yaml
```diff
@@ -149,7 +149,7 @@ scipy_version:
 setuptools_version:
   - '>=49,<50'
 spdlog_version:
-  - '>=1.8.5,<2.0.0a0'
+  - '=1.8.5'
```
jakirkham, line 152: What about allowing patch fixes?

```suggestion
  - '>=1.8.5,<1.9'
```

cjnolet: Funny, that's what I had at first but ended up being conservative just since we know 1.8.5 works. I'm not at all opposed to changing it, though.

jakirkham: Would suggest to keep things a bit looser as very tight pinnings tend to cause issues of their own (not getting needed bug fixes without significant effort for example)

jakirkham (approved):

dantegd:
@rapidsai/ops-codeowners the PR in RMM has been merged, can this PR be approved/merged to unblock cuML?

ajschmidt8 (approved):

PR #325, opened 2021-07-23
Add cuda11.4

Ethyling:


ajschmidt8 (approved):

Ethyling:
rerun tests

ajschmidt8:
Looks like we have a circular dependency here. The tests here failed because there's no `gpuci/miniconda-cuda-driver:11.4-devel-ubuntu16.04` image, but those images won't publish until we get the `11.4` `rapids-notebook-env` packages published.

I think one of these PRs will need to be force merged for now, but ultimately we should try to undo this circular dependency.

Failure log:
https://gpuci.gpuopenanalytics.com/job/rapidsai/job/integration/job/prb/job/meta-pkg-nightly-build/691/CONDA_CONFIG_FILE=conda%2Frecipes%2Fversions.yaml,CONDA_USERNAME=rapidsai-nightly,CUDA_VER=11.4,PYTHON_VER=3.7,RAPIDS_VER=21.08.00a/console

Ethyling:
> Looks like we have a circular dependency here. The tests here failed because there's no `gpuci/miniconda-cuda-driver:11.4-devel-ubuntu16.04` image, but those images won't publish until we get the `11.4` `rapids-notebook-env` packages published.
> 
> I think one of these PRs will need to be force merged for now, but ultimately we should try to undo this circular dependency.
> 
> Failure log:
> https://gpuci.gpuopenanalytics.com/job/rapidsai/job/integration/job/prb/job/meta-pkg-nightly-build/691/CONDA_CONFIG_FILE=conda%2Frecipes%2Fversions.yaml,CONDA_USERNAME=rapidsai-nightly,CUDA_VER=11.4,PYTHON_VER=3.7,RAPIDS_VER=21.08.00a/console

I think it's expected to not have `gpuci/miniconda-cuda-driver:11.4-devel-ubuntu16.04` as we dropped the support for `ubuntu16.04`. I updated the jobs to use `centos7` base images instead. This should fix the CI

ajschmidt8:
> I think it's expected to not have `gpuci/miniconda-cuda-driver:11.4-devel-ubuntu16.04` image as we dropped the support for `ubuntu16.04`. I updated the jobs to use `centos7` based images instead. This should fix the CI

Good call. I was looking at the cuda version instead of the OS. It look like we do have `11.4` images published for centos7 :+1:

Ethyling:
But on a side note, I still think there is a circular dependency. So to add a new version of cuda, we first need to add it only for `gpuci/miniconda-cuda-driver`, then build those packages, and then add it for all the `gpuci` images pipeline. it worked for `11.3` and `11.4` because the toolkit was downgraded during the installation of the `rapids-env` packages, and I don't think we want that

ajschmidt8:
> But on a side note, I still think there is a circular dependency. So to add a new version of cuda, we first need to add it only for `gpuci/miniconda-cuda-driver`, then build those packages, and then add it for all the `gpuci` images pipeline. it worked for `11.3` and `11.4` because the toolkit was downgraded during the installation of the `rapids-env` packages, and I don't think we want that

Yeah, agreed. It seems that the `gpuci/rapidsai` image build stage of the `gpuci-build-scripts` repo pipeline depends on the `rapids-notebook-env` package here, but the `integration` repo here depends on the `gpuci/miniconda-cuda-driver` image which can't publish correctly if the `gpuci/rapidsai` fails to build.

Does that seem accurate to you @Ethyling?

Ethyling:
> > But on a side note, I still think there is a circular dependency. So to add a new version of cuda, we first need to add it only for `gpuci/miniconda-cuda-driver`, then build those packages, and then add it for all the `gpuci` images pipeline. it worked for `11.3` and `11.4` because the toolkit was downgraded during the installation of the `rapids-env` packages, and I don't think we want that
> 
> Yeah, agreed. It seems that the `gpuci/rapidsai` image build stage of the `gpuci-build-scripts` repo pipeline depends on the `rapids-notebook-env` package here, but the `integration` repo here depends on the `gpuci/miniconda-cuda-driver` image which can't publish correctly if the `gpuci/rapidsai` fails to build.
> 
> Does that seem accurate to you @Ethyling?

Yes. That's why I think that when we add a new cuda version, we should proceed in two steps for the `gpuci-build-environment` repo. First we add the new version just for `gpuci/miniconda-cuda-driver` and the previous images in the pipeline, and then we add the new version for all the images in the pipeline once we have the `rapids-env` packages available for this new version

Ethyling:
It seems we will need `xgboost` for `cuda11.4` before merging this

Ethyling:
rerun tests

raydouglass (approved):

PR #346, opened 2021-09-09
Add jupyterlab-favorites to notebook env

jacobtomlinson:
When starting Jupyter Lab you need to decide which directory the file explorer starts in. In some RAPIDS deployment environments it is not always obvious which directory you should choose. Should it be `/rapids`, or `$HOME`? There may also be datasets or workspaces mounted in HPC/cloud environments which may be the actual desired workspace.

Today I came across this nice small Jupyter Lab extension from NERSC. It adds a favourite directories section at the top of the file explorer in Jupyter Lab and allows users to add and remove directories. 

image

It seems particularly useful because an admin can drop in a config ahead of time with some useful directories. I intend to raise a follow-up PR to rapidsai/docker to add `/rapids` and `$HOME` to this default list.

### DIFF
--- a/conda/recipes/rapids-notebook-env/meta.yaml
+++ b/conda/recipes/rapids-notebook-env/meta.yaml
@@ -51,6 +51,7 @@ requirements:
     - ipywidgets
     - jupyter-server-proxy
     - jupyterlab {{ jupyterlab_version }}
+    - jupyterlab-favorites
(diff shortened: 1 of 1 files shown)


ajschmidt8 (approved):

jacobtomlinson:
Thanks @ajschmidt8

PR #370, opened 2021-10-01
Change `cudf` test directory

ajschmidt8:
This PR updates the `cudf` test directory to `/rapids/cudf/python/cudf/cudf/tests` to ensure that it excludes the `/rapids/cudf/python/cudf/cudf/benchmarks` directory, which is not currently working and does not need to be tested in CI.

Additionally, it changes the `py.test` command to `pytest` since `py.test` is considered deprecated ([src](https://stackoverflow.com/a/41893170)).


msadang (approved):
Look good.  No additional comments.

PR #375, opened 2021-10-11
Add arm64 release axis

jjacobelli:
We will do a copy of this [job](https://gpuci.gpuopenanalytics.com/job/rapidsai/job/integration/job/release/job/main-release-build/) for `arm64` and use this new axis file for it

### DIFF
--- a/ci/axis/release-arm64.yaml
+++ b/ci/axis/release-arm64.yaml
@@ -1,0 +1,20 @@
+CONDA_USERNAME:
+  - rapidsai
+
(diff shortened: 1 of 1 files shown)


Ethyling:
> Is there an associated Jenkins job? If so, can you post it here for reference? Ideally this should go in the PR body. We should try to avoid leaving the PR bodies blank.

We will do a copy of this job for `arm64`. Should we add a job on top of `amd64` and `arm64` jobs?

PR #400, opened 2021-11-29
updated yaml files to remove 3.7 and include 3.9

msadang:
Update python versions to only use 3.8 and 3.9 and remove 3.7.

### DIFF
--- a/ci/axis/tests.yaml
+++ b/ci/axis/tests.yaml
@@ -21,6 +21,6 @@ LINUX_VER:
   - centos8
 
 PYTHON_VER:
-  - 3.7
+  - 3.9
 
 exclude:
--- a/ci/axis/nightly-arm64.yaml
+++ b/ci/axis/nightly-arm64.yaml
@@ -15,7 +15,7 @@ CUDA_VER:
   - 11.2
 
 PYTHON_VER:
-  - 3.7
+  - 3.9
   - 3.8
 
 exclude:
--- a/ci/axis/nightly.yaml
+++ b/ci/axis/nightly.yaml
@@ -16,7 +16,7 @@ CUDA_VER:
   - 11.0
 
 PYTHON_VER:
-  - 3.7
+  - 3.9
   - 3.8
 
 exclude:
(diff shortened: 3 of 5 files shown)


msadang:
https://github.com/rapidsai/ops/issues/1280

ci/axis/tests.yaml
```diff
@@ -21,6 +21,6 @@ LINUX_VER:
   - centos8
 
 PYTHON_VER:
-  - 3.7
+  - 3.9
```
ajschmidt8, line 24:
```suggestion
  - 3.8
```

Let's switch this to `3.8` to continue testing on the oldest version of Python that we support.

ajschmidt8 (changes requested):

ci/axis/tests.yaml
msadang: Change completed

msadang:
rerun tests

conda/recipes/versions.yaml
```diff
@@ -15,7 +15,7 @@ xgboost_version:
 conda_version:
   - '=4.10.3'
 conda_build_version:
-  - '=3.20.3'
+  - '=3.21.7'
```
ajschmidt8, line 18: Is there a reason the `conda_build_version` variable is being updated in this PR?

msadang: conda_build caused a build error.  3.20.3 doesn't have a python 3.9 version.

Ethyling:
It seems like we need to upgrade `ipython` to a newer version as version `7.15` does not have python 3.9 packages: https://anaconda.org/conda-forge/ipython/files?version=7.15.0

Ethyling:
We also need to update the PR to match the changes made by https://github.com/rapidsai/integration/pull/406
We should only add python 3.9 in the `nightly-env-arm64.yaml` and `nightly-env.yaml` files

msadang:
rerun tests

Ethyling:
rerun tests

ajschmidt8 (approved):

PR #406, opened 2022-01-10
Split axis files for env and meta packages

Ethyling:
Currently we are using the same axis files for `env` and `meta` packages. Due to this, when we add a new cuda or python version in the axis files, the CI for `meta` packages is failing as rapids packages are not yet available for this new version. This PR split these files for `env` and `meta` packages.

We also need to update the CI jobs to use the new axis files.


ajschmidt8 (approved):

Ethyling:
rerun tests

Ethyling:
Admin merging as CI failure is not related to this PR

PR #414, opened 2022-01-17
allow scikit-image 0.19 in the conda environment

grlee77:
This PR is to allow scikit-image 0.19 in the test environment as requested in https://github.com/rapidsai/cucim/pull/190#pullrequestreview-848290209. It is not strictly necessary to update to 0.19 in order for the tests in that PR to pass (the scikit-image package is mainly used to retrieve example data that is used in the tests for cuCIM and all the datasets we use are already present in 0.18)


jakirkham:
cc @ajschmidt8 @Ethyling @gigony

jakirkham:
Thanks all! 😀

PR #435, opened 2022-02-18
Drop Python 3.7 requirements

jakirkham:
As we've dropped Python 3.7, we can also drop the requirements used with 3.7.


msadang (approved):
LGTM

PR #453, opened 2022-03-29
Pin click version to last supported by black 22.1.0.

vyasr:
Our 22.04 will continue using black < 22.3.0, so we need the appropriate Click pinning to make sure that black continues to work.


vyasr:
I think this can safely be admin-merged. The failures are unrelated (we're already handling the cuxfilter issue elsewhere, not sure what's going on with the mnmg tests).

galipremsagar (approved):

PR #457, opened 2022-03-30
Remove now unnecessary click pinning.

vyasr:
Removes the pinning introduced in #453 on branch 22.06 since that pinning was specifically needed only for branch-22.04.


ajschmidt8:
relevant tests have passed. admin merging.

PR #468, opened 2022-04-21
Added myst-nb and sphinx-autobuild to rapids-docs-env

mmccarty:
Adding dependencies for rapidsai/cudf#10685

### DIFF
--- a/conda/recipes/rapids-doc-env/meta.yaml
+++ b/conda/recipes/rapids-doc-env/meta.yaml
@@ -51,6 +51,8 @@ requirements:
     - sphinx-click
     - sphinx-copybutton
     - sphinx-markdown-tables
+    - sphinx-autobuild
+    - myst-nb
 
 about:
   home: http://rapids.ai/


conda/recipes/rapids-doc-env/meta.yaml
```diff
@@ -51,6 +51,8 @@ requirements:
     - sphinx-click
     - sphinx-copybutton
     - sphinx-markdown-tables
+    - sphinx-autobuild
+    - myst-nb
```
ajschmidt8, line 55: @mmccarty, is this `myst-nb` dependency for the notebooks? If so, we should probably put it in the `rapids-notebook-env` recipe below instead of here.

- https://github.com/rapidsai/integration/blob/branch-22.06/conda/recipes/rapids-notebook-env/meta.yaml

mmccarty: Yes, but only for executing and rendering the notebooks contained with in the docs (like [here](https://github.com/rapidsai/cudf/tree/branch-22.06/docs/cudf/source/user_guide)). In other words, we don't need `myst-nb` outside of `sphinx`.

ajschmidt8 (approved):

PR #469, opened 2022-04-22
Fix integration tests

ajschmidt8:
This PR removes a step to install `requests`. `requests` is already available in our images and this step seems to be causing `conda` solve issues.

Ultimately these integration tests should eventually be removed from this repository, but this should at least get them working again for now.


Ethyling (approved):

ajschmidt8:
relevant tests have passed. admin merging.

PR #484, opened 2022-06-06
Switch `rapids-build-env` to `conda` compilers

ajschmidt8:
This PR ensures that the `rapids-build-env` package uses `conda` compilers that are similar to the ones used in our recent recipe conversions.

i.e.:

- https://github.com/rapidsai/cudf/pull/10275
- https://github.com/rapidsai/cuml/pull/4577


conda/recipes/rapids-build-env/meta.yaml
```diff
@@ -59,6 +61,7 @@ requirements:
     - cuda-python {{ cuda_python_version }}
     - cudatoolkit ={{ cuda_major }}.*
     - cupy {{ cupy_version }}
+    - cxx-compiler {{ cxx_compiler_version }}
```
jakirkham, line 64: Would recommend including the `c-compiler` too

Also the compiler versions are [a bit wonky]( https://github.com/conda-forge/compilers-feedstock/blob/084a965125ed37df62f944c027ca6d64806b35e8/recipe/meta.yaml#L29 ). So would skip pinning these directly

Though we can do this indirectly by pinning `gcc_impl_*`

```suggestion
    - c-compiler
    - cxx-compiler
    - gcc_impl_{{ target_platform }}  {{ gcc_version }}
```

conda/recipes/versions.yaml
```diff
@@ -19,9 +19,11 @@ conda_build_version:
 conda_verify_version:
   - '=3.1.1'
 
-# Versions for `libgcc-ng`, `libgfortran-ng`, `libstdcxx-ng` stack
-build_stack_version:
-  - '>=9.4.0'
+# conda compilers
+cxx_compiler_version:
+  - 9
```
jakirkham, line 24:
```suggestion
gcc_version:
  - 9
```

jakirkham (commented):
Thanks AJ! 🙏

Generally this looks good. Had a few minor comments about pinning compilers below

jakirkham (approved):

sevagh (approved):

PR #529, opened 2022-09-03
Update to XGBoost 1.6.2

jakirkham:
cc @mmccarty (for awareness)

PR #528, opened 2022-08-25
Upgrade to cupy 11

galipremsagar:
Required for https://github.com/rapidsai/cudf/pull/11449#pullrequestreview-1085735332

PR #530, opened 2022-09-06
Fix CI script

ajschmidt8:
This PR makes two changes to our CI script:

- Adds a fix for the issue that's currently plaguing #528 and #529
- Removes a config value that is no longer needed since we removed the `conda` mirror


AyodeAwe (approved):
LGTM

PR #533, opened 2022-09-14
Trying to resolve a version issue with GDAL and TensorFlow, cuXFilter, etc.

thomcom:
cuspatial has GDAL pinned to 3.4.x which is causing a conflict with https://github.com/rapidsai/cuspatial/issues/674 . I'm updating here and cuspatial to use 3.*. Contributes to https://github.com/rapidsai/cuspatial/issues/674

### DIFF
--- a/conda/recipes/versions.yaml
+++ b/conda/recipes/versions.yaml
@@ -71,7 +71,7 @@ flake8_version:
 fsspec_version:
   - '>=0.9.0'
 gdal_version:
-  - '>=3.0.0'
+  - '>=3.5.0'
 geopandas_version:
   - '>=0.11.0'
 gcsfs_version:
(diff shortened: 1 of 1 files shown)


ajschmidt8:
@thomcom, we need to keep the following CVE in mind when editing the GDAL version. The affected GDAL versions are `3.3.0-3.4.0`. Ideally we'd bump to newer versions. Some of the older versions caused conda conflicts in our images (see https://github.com/rapidsai/integration/pull/521 and https://github.com/rapidsai/cuspatial/pull/623)

- https://nvd.nist.gov/vuln/detail/CVE-2021-45943

thomcom:
Thanks for that reminder @ajschmidt8 I think that we have GDAL pinned specifically to `3.4.3`

thomcom:
I'm still experimenting with a good GDAL version to support, but it'll probably be set to `3.5` which is what I was going with on this one.

ajschmidt8 (approved):

PR #546, opened 2022-10-10
Fix `update-version.sh` with correct files

raydouglass:
Updates the files that get updated by `update-version.sh`.

Also adds a workaround for MacOS BSD sed to use GNU `gsed` instead. BSD sed gives the following error:
```
sed: 1: "/RAPIDS_VER/{n; s/- .*/ ...": bad flag in substitute command: '}'
```


raydouglass:
Failures are unrelated to the change, admin merging.

PR #554, opened 2022-11-08
Add `versioneer` to build dependencies

jakirkham:
As it is easier to upgrade `versioneer` when it is devendored and it works better with `pyproject.toml`, add `versioneer` as a build dependency to service libraries moving in this direction.


msadang (approved):
Version changes look good.

PR #598, opened 2023-01-31
Adds CUDA 11.8 & Python 3.10 to meta packages

raydouglass:
Adds CUDA 11.8 & Python 3.10 to meta packages. Tests are not yet updated since they run on `rapidsai/rapidsai-core-nightly` which doesn't have CUDA 11.8 until after the meta packages are built.

### DIFF
--- a/ci/axis/nightly-env-arm64.yaml
+++ b/ci/axis/nightly-env-arm64.yaml
@@ -14,7 +14,6 @@ CUDA_VER:
 
 PYTHON_VER:
   - 3.8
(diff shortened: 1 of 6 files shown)


raydouglass:
> Should we also remove `3.9` now?

Good question. I was thinking that since all the libraries are still building Python 3.9, this repo should too and we would remove it in a follow up PR.

raydouglass:
On second thought, I think I will remove Python 3.9.

PR #607, opened 2023-02-03
Update faiss version to match cuml

raydouglass:
Update to match `cuml`: https://github.com/rapidsai/cuml/blob/branch-23.02/dependencies.yaml#L135

### DIFF
--- a/conda/recipes/versions.yaml
+++ b/conda/recipes/versions.yaml
@@ -63,7 +63,7 @@ double_conversion_version:
 doxygen_version:
   - '>=1.8.12,<=1.8.20'
 faiss_version:
-  - '>=1.7.1'
+  - '=1.7.2'
 fastavro_version:
   - '>=0.22.0'
 flake8_version:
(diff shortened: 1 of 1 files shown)


AyodeAwe (approved):

raydouglass:
Dante pointed out that `cuml` actually inherits `raft`'s `faiss` version here: https://github.com/rapidsai/raft/blob/branch-23.02/conda/recipes/libraft/conda_build_config.yaml#L23

ajschmidt8 (approved):

bdice:
We need to pin this to `"1.7.2 *_cuda"`. The GPU selector package won't work properly with channel priority and we have to pull faiss from `rapidsai`. The conda-forge variants don't have ARM+CUDA at all, and won't support x86_64 on arch 90. We built packages of libfaiss for the `rapidsai` channel from the recipe in this PR: https://github.com/conda-forge/faiss-split-feedstock/pull/64

I will open a new PR to update that pinning.

PR #609, opened 2023-02-03
Pin two notebook-env packages

raydouglass:
Temporary solution for https://github.com/jupyter-server/jupyter_server/issues/1198

### DIFF
--- a/conda/recipes/versions.yaml
+++ b/conda/recipes/versions.yaml
@@ -86,8 +86,14 @@ holoviews_version:
   - '>=1.14.8,<=1.15.3'
 ipython_version:
   - '=7.31.1'
+# TODO Update ipykernel_version once https://github.com/jupyter-server/jupyter_server/issues/1198 is resolved
+ipykernel_version:
+  - '=6.20.2'
 isort_version:
   - '=5.12.0'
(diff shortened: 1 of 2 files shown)


conda/recipes/versions.yaml
```diff
@@ -86,8 +86,14 @@ holoviews_version:
...
 ipython_version:
   - '=7.31.1'
+# TODO Update ipykernel_version once https://github.com/jupyter-server/jupyter_server/issues/1198 is resolved
+ipykernel_version:
+  - '=6.20.2'
 isort_version:
   - '=5.12.0'
+# TODO Update jupyter_core_version once https://github.com/jupyter-server/jupyter_server/issues/1198 is resolved
+jupyter_core_version:
+  - '=5.1.5'
```
pentschev, line 96: From my testing this seemed unnecessary, just `ipykernel` was needed. I doesn't seem to harm either though, so I'll leave it up to you.

ajschmidt8 (approved):

conda/recipes/versions.yaml
raydouglass: I think I want to be overly cautious here since we are so close to the release.

PR #615, opened 2023-02-06
Fix `test.yaml` axis

ajschmidt8:
This axis control which `devel` containers get tested on a nightly basis.

Since we've recently reduced our `devel` container matrix, this axis file should be updated to reflect those changes.

### DIFF
--- a/ci/axis/tests.yaml
+++ b/ci/axis/tests.yaml
@@ -9,15 +9,12 @@ RAPIDS_VER:
   - "23.02"
 
 CUDA_VER:
-  - 11.5
+  - 11.8
 
 LINUX_VER:
-  - ubuntu18.04
-  - ubuntu20.04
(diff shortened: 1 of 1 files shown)


ci/axis/tests.yaml
```diff
@@ -9,15 +9,12 @@ RAPIDS_VER:
...
 LINUX_VER:
-  - ubuntu18.04
-  - ubuntu20.04
-  - centos7
-  - centos8
+  - ubuntu22.04
 
 PYTHON_VER:
-  - 3.8
+  - 3.10
```
raydouglass, line 18: Need to quote this, since yaml will parse it as `3.1`.

raydouglass (changes requested):

raydouglass (approved):

PR #618, opened 2023-02-09
Fix `dask` pins

ajschmidt8:
This PR updates the `dask` and `distributed` pins to match our libraries (e.g. [here](https://github.com/rapidsai/cuml/blob/branch-23.04/conda/recipes/cuml/meta.yaml#L71-L72)).

When only a single `=` sign is used, it is equivalent to `2023.1.1.*`, which causes `conda` to pick up newer nightly packages like `2023.1.1a230127`.

If a newer nightly package gets installed in our base images, it prevents our end-user images from picking up the latest RAPIDS packages which have the `==` pinnings for `dask`.

cc: @galipremsagar

### DIFF
--- a/conda/recipes/rapids-build-env/meta.yaml
+++ b/conda/recipes/rapids-build-env/meta.yaml
@@ -63,6 +63,7 @@ requirements:
     - gcc_impl_{{ target_platform }}  {{ gcc_version }}
     - cython {{ cython_version }}
     - dask {{ dask_version }}
+    - dask-core {{ dask_version }}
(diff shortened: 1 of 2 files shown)


galipremsagar (approved):
TIL, Thanks @ajschmidt8 !

raydouglass (approved):

galipremsagar:
I just went back and verified in all the dask pinning PRs, everything else looks clean.

ajschmidt8:
Thanks prem!

PR #623, opened 2023-02-13
Update ipykernel pin to reject broken multiprocessing releases

pentschev:
As per https://github.com/jupyter-server/jupyter_server/issues/1198 ipykernel version 6.21.0, 6.21.1 were broken when running with multiprocessing module. As of 6.21.2 this is now fixed by https://github.com/ipython/ipykernel/pull/1095 .


raydouglass (approved):

raydouglass:
The test error is due to a version mismatch between `cugraph` and `libcugraph`, so I will admin merge.

PR #627, opened 2023-03-06
Update minimum `pandas` and `numpy` pinnings

galipremsagar:
This PR:

 Increments the minimum pinning for pandas version from `1.0` to `1.3`.
 Sets a minimum pinning for `numpy` as `>=1.21`
 Fixes arm conda environment creation by removing `pandoc` version constraint.

xref: https://github.com/rapidsai/cudf/pull/12887

### DIFF
--- a/conda/recipes/rapids-doc-env/meta.yaml
+++ b/conda/recipes/rapids-doc-env/meta.yaml
@@ -43,7 +43,7 @@ requirements:
     - markdown {{ markdown_version }}
     - nbsphinx {{ nbsphinx_version }}
     - numpydoc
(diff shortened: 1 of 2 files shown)


conda/recipes/rapids-doc-env/meta.yaml
```diff
@@ -1,5 +1,6 @@
 # Copyright (c) 2020-2023, NVIDIA CORPORATION.
 
+
```
galipremsagar, line 3:
```suggestion
```

raydouglass (approved):

PR #642, opened 2023-05-05
Add GHAs workflows

ajschmidt8:
This PR adds GitHub Action workflows to the `integration` repository.

I will remove the remaining Jenkins scripts in a subsequent PR.


ajschmidt8:
In Jenkins, the `env` and `meta` packages were split up, but I don't think this is necessary for the GHAs work since the `env` packages will be going away once the end-user Docker overhaul is complete anyway.

AyodeAwe (approved):

PR #643, opened 2023-05-08
Add `conda-pack` job

ajschmidt8:
This PR adds a `conda-pack` job to the `build.yaml` workflow, which I forgot to include in #642.

It also:
- updates the environment variables to reference vars in the new GHAs CI images
- adds some entries to `update-version.sh`

### DIFF
--- a/ci/conda-pack.sh
+++ b/ci/conda-pack.sh
@@ -1,15 +1,15 @@
 #!/bin/bash
 set -e
 
-RAPIDS_VER="23.06a"
+RAPIDS_VER="23.06"
+VERSION_DESCRIPTOR="a"
 CONDA_USERNAME="rapidsai-nightly"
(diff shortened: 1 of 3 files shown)


ci/conda-pack.sh
```diff
@@ -1,7 +1,15 @@
 #!/bin/bash
 set -e
 
-CONDA_ENV_NAME="rapids${RAPIDS_VER}_cuda${CUDA_VER}_py${PYTHON_VER}"
+RAPIDS_VER="23.06a"
```
raydouglass, line 4: I think this will fail on the `mamba create`. I think it can just be `23.06` and use the `CONDA_USERNAME` to control whether it pulls from nightly or release.

raydouglass (approved):

PR #535, opened 2022-09-28
Upgrade minimum numba version to `0.56.2`

galipremsagar:
Required for https://github.com/rapidsai/cudf/pull/11701/

PR #648, opened 2023-05-26
Update to numba>=0.57.

bdice:
cuDF now requires numba>=0.57.

I am not sure if this change is strictly needed or if cuDF's requirements will automatically enforce the greater lower bound.


gmarkall (approved):
LGTM - I don't know so much about the use of this repo / tooling, but I note it follows the same pattern as the merged https://github.com/rapidsai/integration/pull/535 and I think generally ensuring the Numba version is 0.57 is 👍

raydouglass (approved):

PR #667, opened 2023-07-25
Use cuda-version instead of cudatoolkit for conda pack

raydouglass:
Just use the newer `cuda-version` to select the CUDA version. This will pull in the appropriate packaged based on whether it is CUDA 11 or 12.

Fixes: https://github.com/rapidsai/integration/actions/runs/5661037594/job/15338557625#step:7:48


jakirkham (approved):
Thanks Ray! 🙏

ajschmidt8 (approved):

PR #666, opened 2023-07-25
Remove the env packages

raydouglass:
Removes:
- `rapids-build-env`
- `rapids-doc-env`
- `rapids-notebook-env`

Also updates the `version.yaml` files to remove unnecessary packages.

PR #668, opened 2023-07-25
Always use `cuda-version` in recipes

jakirkham:
As `cuda-version` is supported on all CUDA versions, use it constrain CUDA dependence. Only use `cudatoolkit` on CUDA 11 where it is available and not on CUDA 12 where it is gone.


bdice:
@jakirkham This conflicts with #666, I think. I didn't update those `-env` packages in #662 because they are slated for removal once the new Docker images were released.

jakirkham:
Ah missed that. No strong feelings on approach here. Will defer to @raydouglass

ajschmidt8:
Yeah, this PR can be closed. These packages will be removed before the `23.08` release.

Closing.

PR #664, opened 2023-07-25
Temporarily disable cucim and xgboost in CUDA 12 rapids builds.

bdice:
Discussed with @raydouglass @vyasr @jakirkham -- temporarily disabling `cucim` and `rapids-xgboost` requirements in `rapids` until their corresponding issues are complete:
- https://github.com/rapidsai/cucim/issues/513
- https://github.com/rapidsai/xgboost-feedstock/issues/4

PR #669, opened 2023-07-25
Re-enable cucim and xgboost in CUDA 12 rapids builds.

bdice:
PR #664 temporarily disabled CUDA 12 packages for cucim and xgboost in `rapids`. This re-enables those.

This reverts commit cc272c465fc8904737e284a3061b5284d2f3e054.

This can be merged once the following issues are closed:
- https://github.com/rapidsai/cucim/issues/513
- https://github.com/rapidsai/xgboost-feedstock/issues/4


bdice:
I also enabled testing of conda packages. I think that even if we can't import GPU libraries, we should be able to ensure that the packages contained in `rapids` or `rapids-xgboost` are _solvable_.

conda/recipes/rapids-xgboost/meta.yaml
```diff
@@ -28,13 +28,14 @@ requirements:
...
   run:
-    - {{ pin_compatible('cuda_version', max_pin='x', min_pin='x') }}
+    - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
     {% if cuda_major == "11" %}
     - cudatoolkit
     {% endif %}
     - nccl {{ nccl_version }}
     - python
-    - xgboost {{ xgboost_version }}{{ major_minor_version }}
+    - xgboost {{ xgboost_version }}
```
jakirkham, line 37: Think we need this to ensure we get our flavor of `xgboost` build for CUDA

```suggestion
    - xgboost {{ xgboost_version }} rapidsai_cuda*
```

conda/recipes/rapids-xgboost/meta.yaml
```diff
@@ -28,13 +28,14 @@ requirements:
...
+    - _py-xgboost-mutex=*=rapidsai*
```
jakirkham, line 38: Think we can drop this as the `build/string` constraint above should capture this same information

```suggestion
```

conda/recipes/versions.yaml
```diff
@@ -1,7 +1,7 @@
 # Versions for `rapids-xgboost` meta-pkg
 xgboost_version:
   # Minor version is appended in meta.yaml
-  - '=1.7.5dev.rapidsai'
+  - '=1.7.4'
```
jakirkham, line 4: For context we were building dev versions before (and applying patches). Those patches were largely upstreamed (though we still have a couple patches)

However we now build a stable version of XGBoost (instead of from a commit as before). Hence 1.7.4 is reasonable

jakirkham (commented):
Thanks Bradley! 🙏

Made a few minor comments below

conda/recipes/versions.yaml
bdice: Thanks!

bdice:
We're going to have to disable tests for `rapids` since it depends on `rapids-xgboost` which is built separately. I would like to fix this in a follow-up. For now, testing on `rapids-xgboost` found a bug that I fixed in d1aab9c. I am currently testing locally with `rapids-xgboost` removed from `rapids` (so that it can solve) but I am facing difficulty with solving `custreamz`. I thought this would be fixed by https://github.com/rapidsai/cudf/pull/13754 but mamba doesn't seem to be finding the new packages...

jakirkham:
Updating this thread, sounds like we need PR ( https://github.com/rapidsai/cudf/pull/13769 ) to fix a couple more `build/string`s

jakirkham:
That PR has been merged and packages uploaded

ci/build_python.sh
```diff
@@ -12,12 +12,13 @@ rapids-print-env
 rapids-logger "Build rapids-xgboost"
 
 rapids-mamba-retry mambabuild \
-  --no-test \
   --variant-config-files "${CONDA_CONFIG_FILE}" \
```
jakirkham, line 15:
```suggestion
  --use-local \
  --variant-config-files "${CONDA_CONFIG_FILE}" \
```

ci/build_python.sh
```diff
@@ -12,12 +12,13 @@ rapids-print-env
...
   conda/recipes/rapids-xgboost
 
 rapids-logger "Build rapids"
 
+# TODO: --no-test is enabled because this depends on rapids-xgboost from the
+# above recipe. These should be built as a single recipe to avoid that problem.
 rapids-mamba-retry mambabuild \
   --no-test \
```
jakirkham, line 23:
```suggestion
  --use-local \
```

jakirkham (commented):
Thanks Bradley! 🙏

Think we may need `--use-local` to pick up locally built packages when testing

ci/build_python.sh
```diff
@@ -12,12 +12,13 @@ rapids-print-env
...
 
 rapids-mamba-retry mambabuild \
-  --no-test \
   --variant-config-files "${CONDA_CONFIG_FILE}" \
   conda/recipes/rapids-xgboost
 
 rapids-logger "Build rapids"
 
+# TODO: --no-test is enabled because this depends on rapids-xgboost from the
+# above recipe. These should be built as a single recipe to avoid that problem.
```
jakirkham, line 21:
```suggestion
```

raydouglass (approved):

jakirkham:
Seeing this [on CI]( https://github.com/rapidsai/integration/actions/runs/5674080526/job/15376956190?pr=669#step:7:493 ):

```
The reported errors are:
- Encountered problems while solving:
-   - nothing provides __cuda needed by xgboost-1.7.4-rapidsai_cuda112py38hf635370_1
```

Think we need to set [the environment variable `CONDA_OVERRIDE_CUDA`]( https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-virtual.html ) to some value to "convince" Conda it is ok to install

ci/build_python.sh
```diff
@@ -12,14 +12,14 @@ rapids-print-env
 rapids-logger "Build rapids-xgboost"
```
jakirkham, line 13: Though would suggest we put this up near `CONDA_CONFIG_FILE` above

```suggestion

CONDA_OVERRIDE_CUDA="${RAPIDS_CUDA_VERSION}"

```

jakirkham:
Should we add this to both of these (as we've done with other recipes [like cuDF]( https://github.com/rapidsai/cudf/blob/2231b15c9f256c4885dc28e27e2767b2861290d6/conda/recipes/cudf/meta.yaml#L105-L107 ))

```yaml
test:
  requires:
    - cuda-version ={{ cuda_version }}
```

ci/build_python.sh
```diff
@@ -6,20 +6,21 @@ set -euo pipefail
 source rapids-env-update
 
 CONDA_CONFIG_FILE="conda/recipes/versions.yaml"
+CONDA_OVERRIDE_CUDA="${RAPIDS_CUDA_VERSION}"
```
jakirkham, line 9:
```suggestion
export CONDA_OVERRIDE_CUDA="${RAPIDS_CUDA_VERSION}"
```

conda/recipes/rapids-xgboost/meta.yaml
```diff
@@ -28,13 +28,13 @@ requirements:
...
   run:
-    - {{ pin_compatible('cuda_version', max_pin='x', min_pin='x') }}
+    - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
     {% if cuda_major == "11" %}
     - cudatoolkit
     {% endif %}
     - nccl {{ nccl_version }}
     - python
-    - xgboost {{ xgboost_version }}{{ major_minor_version }}
+    - xgboost {{ xgboost_version }} rapidsai_cuda*
```
jakirkham, line 38:
```suggestion

test:
  requires:
    - cuda-version ={{ cuda_version }}

```

conda/recipes/rapids/meta.yaml
```diff
@@ -69,23 +61,23 @@ requirements:
...
     - conda-forge::ucx {{ ucx_version }}
 
-test:              # [linux64]
-  imports:         # [linux64]
-    - cucim        # [linux64]
-    - cudf         # [linux64]
-    - cudf_kafka   # [linux64]
-    - cugraph      # [linux64]
-    - cuml         # [linux64]
+test:
```
jakirkham, line 64:
```suggestion
test:
  requires:
    - cuda-version ={{ cuda_version }}
```

bdice:
We’ll need to disable the import tests because this runs on CPU — but just solving the environment with no imports is better than no testing.

vyasr:
Looks like we can't actually run the tests on the runner we're using because it doesn't have a working CUDA installation, but at least testing that the environment solves is helpful.

vyasr:
> We’ll need to disable the import tests because this runs on CPU — but just solving the environment with no imports is better than no testing.

Jinx

ci/build_python.sh
```diff
@@ -6,20 +6,21 @@ set -euo pipefail
 source rapids-env-update
 
 CONDA_CONFIG_FILE="conda/recipes/versions.yaml"
+export CONDA_OVERRIDE_CUDA="${RAPIDS_CUDA_VERSION}"
```
bdice, line 9: We should probably consider the implications here. Today, I think rapids is installable even on CPU-only machines. The new rapids-xgboost package design requires `__cuda` to **install**. This is important to support for cases like HPC systems with CPU login nodes and GPU worker nodes that use the same environment.

conda/recipes/rapids-xgboost/meta.yaml
```diff
@@ -28,13 +28,17 @@ requirements:
...
+
+test:
+  requires:
+    - cuda-version ={{ cuda_version }}
```
jakirkham, line 41:
```suggestion
    - cuda-version ={{ cuda_version }}
  commands:
    - exit 0
```

ci/build_python.sh
jakirkham: Yeah they can also install by setting `CONDA_OVERRIDE_CUDA` to some value

In any event, this is [coming from `libxgboost`]( https://github.com/conda-forge/xgboost-feedstock/blob/829bd2ea851c5b4eddafad846c79e333835a45cb/recipe/meta.yaml#L61 ). So we could move this just to the `rapids-xgboost` if we prefer

conda/recipes/rapids/meta.yaml
```diff
@@ -69,23 +61,11 @@ requirements:
...
-    {% if cuda_major == "11" %}
-    - cusignal     # [linux64]
-    {% endif %}
-    - cuspatial    # [linux64]
-    - custreamz    # [linux64]
-    - cuxfilter    # [linux64]
-    - dask_cuda    # [linux64]
-    - dask_cudf    # [linux64]
-    - pylibcugraph # [linux64]
-    - rmm          # [linux64]
```
jakirkham, line 88: We could run these through [`pkgutil.find_loader`]( https://docs.python.org/3/library/pkgutil.html#pkgutil.find_loader ) in a [`run_test.py` script in the recipe]( https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#run-test-script )

This would let us test for their existence without needing to `import` them (and thus not need a GPU to test)

ci/build_python.sh
vyasr: > The new rapids-xgboost package design 

What is the relevant new change? Is it in libxgboost or in something about how rapids-xgboost is packaged?

bdice: Okay. Well, maybe this requirement already existed. I'm not sure.

The question is whether `__cuda` should be a hard requirement for installation, which is coming from xgboost-related packages. I'm not sure if it was that way for the old xgboost packages we shipped in 23.06 or not. Regardless, it feels funny that no other RAPIDS package has this requirement besides xgboost.

conda/recipes/rapids/meta.yaml
bdice: I don't think this is necessary for this PR, maybe file an issue or PR with this proposal later on. I feel comfortable with the current level of testing, which is higher than what we had before.

jakirkham (approved):
LGTM. Thanks all! 🙏

Had a few comments above, but none of them are blocking

bdice:
Potentially blocking issue with xgboost builds requiring `__cuda`:

Previous release installation did not require `__cuda` to install:

```
$ CONDA_OVERRIDE_CUDA="" mamba create -n rapids-23.06-cpu -c rapidsai -c conda-forge -c nvidia rapids=23.06 python=3.10 cudatoolkit=11.8
```
The above command succeeds.

The proposed changes here would require `__cuda` to install. This was previously not a constraint for users installing `rapids` and poses a challenge for users of (for example) HPC systems where login nodes do not have GPUs, and only worker nodes have GPUs.

```
$ CONDA_OVERRIDE_CUDA="" mamba create -n rapids-23.08-cpu -c rapidsai-nightly -c conda-forge -c nvidia rapids=23.08 python=3.10 cuda-version=12.0 'xgboost=1.7.4*=rapidsai_cuda*'

...

The following package could not be installed
└─ xgboost 1.7.4* rapidsai_cuda* is uninstallable because it requires
   └─ __cuda  , which is missing on the system.
```

The above command fails on CPU-only systems. This is testing the currently nightly `rapids` which doesn't include xgboost, but with xgboost manually added as the recipe does in this PR.

I think this issue is blocking for the 23.08 release, but not necessarily for this PR. I'd be fine with merging this PR and discussing this issue separately. I can revisit this tomorrow with others.

vyasr (approved):

jakirkham:
After discussion offline, we concluded the XGBoost issue is non-blocking for this PR. So it should be good to merge

We are evaluating options to fix XGBoost packages to not require `__cuda` with stakeholders. Once a fix is deployed we can come back and remove the `CONDA_OVERRIDE_CUDA` line to test the fix and simplify things here

ci/build_python.sh
```diff
@@ -6,20 +6,21 @@ set -euo pipefail
...
-  --no-test \
+  --use-local \
   --variant-config-files "${CONDA_CONFIG_FILE}" \
   conda/recipes/rapids-xgboost
 
 rapids-logger "Build rapids"
 
 rapids-mamba-retry mambabuild \
-  --no-test \
+  --use-local \
```
ajschmidt8, line 23: just want to mention that for other repos, we use `--channel "${RAPIDS_CONDA_BLD_OUTPUT_DIR}"`: https://github.com/rapidsai/cudf/blob/abb59c83128f956c7edcb4d7744cb0faecf0026c/ci/build_python.sh#L18-L39

`RAPIDS_CONDA_BLD_OUTPUT_DIR` is set in our CI images: https://github.com/rapidsai/ci-imgs/blob/75cad918c44c6e00480001b24cb764e1b43fa0a5/Dockerfile#L108-L113

It looks like `--use-local` works, I'm just pointing this out in case we want consistency.

raydouglass (approved):

ci/build_python.sh
jakirkham: Dropping this in PR ( https://github.com/rapidsai/integration/pull/673 ), which pulls in the new `xgboost` packages

ci/build_python.sh
jakirkham: Yeah `--use-local` will check the same things. Please see [this list of paths]( https://github.com/conda/conda/blob/dd7373698d1684e8b7d648d730c6a02b5d908edb/conda/base/context.py#L505-L509 ) that Conda checks when `--use-local` is set

PR #670, opened 2023-07-27
Convert CUDA version to major.minor during conda-pack

raydouglass:
Should fix the build problem seen here: https://github.com/rapidsai/integration/actions/runs/5681986332/job/15399831914

Filed an issue to follow up on CI for this: https://github.com/rapidsai/integration/issues/671

### DIFF
--- a/ci/conda-pack.sh
+++ b/ci/conda-pack.sh
@@ -9,7 +9,8 @@ if [ "$GITHUB_REF_TYPE" = "tag" ]; then
     VERSION_DESCRIPTOR=""
     CONDA_USERNAME="rapidsai"
 fi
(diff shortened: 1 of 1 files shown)


ajschmidt8 (approved):
are these files still being used by anyone?

how are they consumed?

I ask because this PR will change the format of the uploaded file name.

and I think when I migrated this job to GHAs, I had changed the file name as well.

so if the filenames are changing and we're not announcing these changes, is anyone actually using these artifacts?

raydouglass:
> are these files still being used by anyone?

They are used infrequently by some RAPIDS team members. There's been some discussions about advertising more widely, but nothing concrete.

PR #676, opened 2023-08-07
Add dask metapackages and documentation

vyasr:
This is currently a proposal for a metapackage. It may be migrated to another repository, but this repo seems like a natural home and is the easiest place to demo it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ See the [README](conda/recipes/README.md) for more information about the metapac
 ## dask Metapackage
 
 This repository provides metapackages for pip and conda that centralize the dask version dependency across RAPIDS.
-dask's API instability means that each RAPIDS release must pin to a very specific dask release to avoid incompatibilities.
+Dask's API instability means that each RAPIDS release must pin to a very specific dask release to avoid incompatibilities.
 These metapackages provide a centralized, versioned storehouse for that pinning.
 The `rapids_dask_dependency` encodes both `dask` and `distributed` requirements.
 
@@ -26,9 +26,9 @@ When RAPIDS hits code freeze and we pin dask versions, the package versions in t
 At this time, a new metapackage will be released, 23.10.00.1.
 This new metapackage version will be automatically picked up by other RAPIDS libraries since they will be using a `==23.10.00.*` pin.
 
-### Requiring dask nightlies
+### Requiring Dask nightlies
 
-Prior to final pinning for a release, dask versions should be specified using PEP 440-compatible versions like `>=2023.7.1a0` so that nightlies may be picked up.
+Prior to final pinning for a release, Dask versions should be specified using PEP 440-compatible versions like `>=2023.7.1a0` so that nightlies may be picked up.
 For conda, nightlies are published to the [dask channel](https://anaconda.org/dask/).
 The metapackage assumes that the `dask/label/dev` channel is included in a user's condarc so that the nightly will be found.
 For pip, no nightlies are published so the packages must be installed directly from source.
@@ -43,4 +43,4 @@ To do so, the metapackage will encode dependencies as:
 If RAPIDS itself requires a patch release, a new metapackage version will be released that bumps the patch version e.g. 23.10.01.0.
 RAPIDS libraries should at this time update their metapackage pinnings to be `==23.10.01.*` so that metapackages corresponding to the patch release are detected.
 Note that patch releases are why we must specify `==` rather than `>=` constraints.
-We do not want a new metapackage release for a RAPIDS patch release to affect lower patch releases, because a patch release of RAPIDS could involve dask changes, necessitating a bump in the dask pinning that we do not want to propagate backwards to the previous patch release.
+We do not want a new metapackage release for a RAPIDS patch release to affect lower patch releases, because a patch release of RAPIDS could involve Dask changes, necessitating a bump in the Dask pinning that we do not want to propagate backwards to the previous patch release.
--- a/README.md
+++ b/README.md
@@ -2,8 +2,45 @@
 
 RAPIDS - combined conda package for all of RAPIDS libraries
 
-## RAPIDS Meta-packages
+## RAPIDS Metapackages
 
-The conda recipe in the `conda` folder provides the RAPIDS meta-packages, which when installed will provide the latest RAPIDS libraries for the given version.
+The conda recipe in the `conda` folder provides the RAPIDS metapackages, which when installed will provide the latest RAPIDS libraries for the given version.
 
-See the [README](conda/recipes/README.md) for more information about the meta-packages and how to update versions.
+See the [README](conda/recipes/README.md) for more information about the metapackages and how to update versions.
+
+## dask Metapackage
+
+This repository provides metapackages for pip and conda that centralize the dask version dependency across RAPIDS.
+dask's API instability means that each RAPIDS release must pin to a very specific dask release to avoid incompatibilities.
+These metapackages provide a centralized, versioned storehouse for that pinning.
+The `rapids_dask_dependency` encodes both `dask` and `distributed` requirements.
+
+### Metapackage Versioning
+
+The package is versioned by adding an extra release segment to the standard RAPIDS CalVer.
+For example, the initial release of the metapackage for 23.10.00 will be 23.10.00.00
+This version of the metapackage will always pull the latest nightlies for the dask and distributed packages.
+RAPIDS repos will only pin up to the RAPIDS patch version, i.e. `==23.10.00.*`.
+
+When RAPIDS hits code freeze and we pin dask versions, the package versions in this repository should be pinned.
+At this time, a new metapackage will be released, 23.10.00.1.
+This new metapackage version will be automatically picked up by other RAPIDS libraries since they will be using a `==23.10.00.*` pin.
+
+### Requiring dask nightlies
+
+Prior to final pinning for a release, dask versions should be specified using PEP 440-compatible versions like `>=2023.7.1a0` so that nightlies may be picked up.
+For conda, nightlies are published to the [dask channel](https://anaconda.org/dask/).
+The metapackage assumes that the `dask/label/dev` channel is included in a user's condarc so that the nightly will be found.
+For pip, no nightlies are published so the packages must be installed directly from source.
+To do so, the metapackage will encode dependencies as:
+```
+- dask @ git+https://github.com/dask/dask.git@main
+- distributed @ git+https://github.com/dask/distributed.git@main
+```
+
+### RAPIDS patch releases
+
+If RAPIDS itself requires a patch release, a new metapackage version will be released that bumps the patch version e.g. 23.10.01.0.
+RAPIDS libraries should at this time update their metapackage pinnings to be `==23.10.01.*` so that metapackages corresponding to the patch release are detected.
+Note that patch releases are why we must specify `==` rather than `>=` constraints.
+We do not want a new metapackage release for a RAPIDS patch release to affect lower patch releases, because a patch release of RAPIDS could involve dask changes, necessitating a bump in the dask pinning that we do not want to propagate backwards to the previous patch release.
--- a/conda/recipes/rapids_dask_dependency/meta.yaml
+++ b/conda/recipes/rapids_dask_dependency/meta.yaml
@@ -1,0 +1,29 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+package:
+- name: rapids_dask_dependency
+  version: 23.10.00.0
+
+build:
+  number: 0
+  noarch: generic
+
+requirements:
+  run:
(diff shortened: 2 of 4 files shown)


galipremsagar (approved):

README.md
```diff
@@ -2,8 +2,45 @@
...
-The conda recipe in the `conda` folder provides the RAPIDS meta-packages, which when installed will provide the latest RAPIDS libraries for the given version.
+The conda recipe in the `conda` folder provides the RAPIDS metapackages, which when installed will provide the latest RAPIDS libraries for the given version.
 
-See the [README](conda/recipes/README.md) for more information about the meta-packages and how to update versions.
+See the [README](conda/recipes/README.md) for more information about the metapackages and how to update versions.
+
+## dask Metapackage
+
+This repository provides metapackages for pip and conda that centralize the dask version dependency across RAPIDS.
+dask's API instability means that each RAPIDS release must pin to a very specific dask release to avoid incompatibilities.
```
pentschev, line 14:
```suggestion
Dask's API instability means that each RAPIDS release must pin to a very specific dask release to avoid incompatibilities.
```

pentschev: Not relevant for the proposal itself, but this is something that truly confused me when I first found it out. The name `dask` (lower-case `d`) is _only_ used in the new logo, while `Dask` (capital `D`) should still be used in writing. See www.dask.org, it always uses `Dask` in writing. I think it's fair to use `dask` when specifically referring to a package (e.g., conda-forge or PyPI).

README.md
```diff
@@ -2,8 +2,45 @@
...
+The package is versioned by adding an extra release segment to the standard RAPIDS CalVer.
+For example, the initial release of the metapackage for 23.10.00 will be 23.10.00.00
+This version of the metapackage will always pull the latest nightlies for the dask and distributed packages.
+RAPIDS repos will only pin up to the RAPIDS patch version, i.e. `==23.10.00.*`.
+
+When RAPIDS hits code freeze and we pin dask versions, the package versions in this repository should be pinned.
+At this time, a new metapackage will be released, 23.10.00.1.
+This new metapackage version will be automatically picked up by other RAPIDS libraries since they will be using a `==23.10.00.*` pin.
+
+### Requiring dask nightlies
```
pentschev, line 29:
```suggestion
### Requiring Dask nightlies
```

README.md
```diff
@@ -2,8 +2,45 @@
...
+
+Prior to final pinning for a release, dask versions should be specified using PEP 440-compatible versions like `>=2023.7.1a0` so that nightlies may be picked up.
```
pentschev, line 31:
```suggestion
Prior to final pinning for a release, Dask versions should be specified using PEP 440-compatible versions like `>=2023.7.1a0` so that nightlies may be picked up.
```

README.md
```diff
@@ -2,8 +2,45 @@
...
+- dask @ git+https://github.com/dask/dask.git@main
+- distributed @ git+https://github.com/dask/distributed.git@main
+```
+
+### RAPIDS patch releases
+
+If RAPIDS itself requires a patch release, a new metapackage version will be released that bumps the patch version e.g. 23.10.01.0.
+RAPIDS libraries should at this time update their metapackage pinnings to be `==23.10.01.*` so that metapackages corresponding to the patch release are detected.
+Note that patch releases are why we must specify `==` rather than `>=` constraints.
+We do not want a new metapackage release for a RAPIDS patch release to affect lower patch releases, because a patch release of RAPIDS could involve dask changes, necessitating a bump in the dask pinning that we do not want to propagate backwards to the previous patch release.
```
pentschev, line 46:
```suggestion
We do not want a new metapackage release for a RAPIDS patch release to affect lower patch releases, because a patch release of RAPIDS could involve Dask changes, necessitating a bump in the Dask pinning that we do not want to propagate backwards to the previous patch release.
```

pip/rapids_dask_dependency/pyproject.toml
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+[project]
+name = "rapids_dask_dependency"
+version = "23.10.00.0"
+description = "Dask and distributed version pinning for RAPIDS"
```
pentschev, line 6:
```suggestion
description = "Dask and Distributed version pinning for RAPIDS"
```

pentschev: `Distributed` can also be capitalized (see the first sentence of https://distributed.dask.org/en/stable/#motivation).

pentschev (changes requested):

README.md
```diff
@@ -2,8 +2,45 @@
...
+### Metapackage Versioning
+
+The package is versioned by adding an extra release segment to the standard RAPIDS CalVer.
+For example, the initial release of the metapackage for 23.10.00 will be 23.10.00.00
+This version of the metapackage will always pull the latest nightlies for the dask and distributed packages.
+RAPIDS repos will only pin up to the RAPIDS patch version, i.e. `==23.10.00.*`.
+
+When RAPIDS hits code freeze and we pin dask versions, the package versions in this repository should be pinned.
+At this time, a new metapackage will be released, 23.10.00.1.
+This new metapackage version will be automatically picked up by other RAPIDS libraries since they will be using a `==23.10.00.*` pin.
```
jakirkham, line 27: Curious whether marking the release as an `alpha` or similar might simplify this release process. Then at release time we could take off the `alpha` and pin

For example with the `dask` nightly conda packages today, we using versioning similar to RAPIDS (like `dask-2023.8.1a230817`)

conda/recipes/rapids_dask_dependency/meta.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+build:
+  number: 0
+  noarch: generic
+
+requirements:
+  run:
+    - dask >=2023.7.1
+    - dask-core >=2023.7.1
+    - distributed >=2023.7.1
```
jakirkham, line 15: As `dask` pulls in the others (and tightly pins them), we can just pin `dask`

If we do want to list these all for completeness, would suggest templating the versions to make things a bit more DRY

```suggestion
    - dask >=2023.7.1
```

conda/recipes/rapids_dask_dependency/meta.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+about:
+  home: https://rapids.ai/
+  license: Custom
+  license_file: conda/recipes/rapids_dask_dependency/LICENSE
```
jakirkham, line 20: The Dask & Distributed licenses are BSD-3-Clause. So we could use the same here for simplicity

Suggesting this particularly if we are shipping this package in our releases (as users may catch this in license auditing tools)

pip/rapids_dask_dependency/pyproject.toml
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+[project]
```
jakirkham, line 3: Should we add a license here too?

conda/recipes/rapids_dask_dependency/meta.yaml
```diff
@@ -0,0 +1,29 @@
...
+- name: rapids_dask_dependency
+  version: 23.10.00.0
+
+build:
+  number: 0
+  noarch: generic
+
+requirements:
+  run:
+    - dask >=2023.7.1
```
jakirkham, line 13: Another question we might want to consider is how tightly we pin. Do we want an upper bound?

pip/rapids_dask_dependency/pyproject.toml
```diff
@@ -0,0 +1,10 @@
...
+name = "rapids_dask_dependency"
+version = "23.10.00.0"
+description = "Dask and distributed version pinning for RAPIDS"
+dependencies = [
+    "dask @ git+https://github.com/dask/dask.git@main",
```
jakirkham, line 8: The equivalent `pip` install to the `conda` install of `dask` would be something like this (please see [`complete`]( https://github.com/dask/dask/blob/c6912aae03eb7ce73ca813e2386c81dbd9c8f119/pyproject.toml#L65-L69 ))

```suggestion
    "dask[complete] @ git+https://github.com/dask/dask.git@main",
```

jakirkham (commented):
Thanks Vyas! 🙏

This has been a long running pain point. Really appreciate you taking time to think about how we can consolidate and simplify Dask usage in RAPIDS 🙂

Generally agree a metapackage seems like the right way to do this

One question that comes to mind is how we version the metapackage. Have noted this below

Noted some other more minor comments as well (though those may be more important when we ship the packages)

README.md
vyasr: The difficulty here will be that pip installs do not pull in pre-releases without either `pip install --pre` or including the alpha tag in the version spec in pyproject.toml. We are planning to migrate RAPIDS wheels to use alphas in the near future, so if we wanted to we could wait on that effort to reach completion before switching to this metapackage. If we did that then we could use alphas here and remove the extra release segment.

jakirkham: FWIW the environment variable `PIP_PRE` works

```bash
$ PIP_PRE=True pip install --dry-run numpy
Collecting numpy
  Obtaining dependency information for numpy from https://files.pythonhosted.org/packages/7f/90/e7829187ad11ea998c26ff69f5581ee7ccaa04f00954829535870c94b07e/numpy-1.26.0b1-cp311-cp311-macosx_11_0_arm64.whl.metadata
  Using cached numpy-1.26.0b1-cp311-cp311-macosx_11_0_arm64.whl.metadata (99 kB)
Using cached numpy-1.26.0b1-cp311-cp311-macosx_11_0_arm64.whl (14.7 MB)
Would install numpy-1.26.0b1
```

<br>

This also works with a `pip.conf`:

```conf
[install]
pre = true
```

```bash
$ pip install --dry-run numpy             
Collecting numpy
  Obtaining dependency information for numpy from https://files.pythonhosted.org/packages/7f/90/e7829187ad11ea998c26ff69f5581ee7ccaa04f00954829535870c94b07e/numpy-1.26.0b1-cp311-cp311-macosx_11_0_arm64.whl.metadata
  Using cached numpy-1.26.0b1-cp311-cp311-macosx_11_0_arm64.whl.metadata (99 kB)
Using cached numpy-1.26.0b1-cp311-cp311-macosx_11_0_arm64.whl (14.7 MB)
Would install numpy-1.26.0b1
```

<br>

Anyways we have some options for configuring `pip` that don't involve altering the command line call directly if that is the concern

<hr>

It's worth noting that conda nightlies for `dask` already need the channel/label `dask/label/dev` added. So there is already a need for passing this alternative options in somehow (this can also be done by altering `condarc` though)

conda/recipes/rapids_dask_dependency/meta.yaml
pentschev: I think keeping them separate still makes sense. Imagine the previous case with Distributed that required us to release a patch release `distributed=7.3.2.1`, whereas `dask=7.3.2`, if we have such a situation in future again this would prevent us from pinning to different patch versions.

conda/recipes/rapids_dask_dependency/meta.yaml
pentschev: During development we never know what the upper bound really is given there's no predictability on Dask features/bugs/release schedule, therefore I think keeping no upper bound as we do currently is more appropriate.

pip/rapids_dask_dependency/pyproject.toml
pentschev: We probably want separate versions here, for the same reasons I previously mentioned in https://github.com/rapidsai/integration/pull/676#discussion_r1299853200

README.md
vyasr: The issue isn't around modifying the command line, it's around the fact that `--pre` automatically propagates to all dependencies. If we can avoid handing users a footgun that could install arbitrary alpha versions into an environment, we should. That being said, the work we're now doing to revamp RAPIDS package versioning should help here.

vyasr: With the other changes in RAPIDS using alphas here should now be feasible.

conda/recipes/rapids_dask_dependency/meta.yaml
vyasr: I'm not sure how licensing a metapackage of ours that just includes other packages (no ours). Should we use a BSD-3-Clause license or an Apache license like the rest of RAPIDS?

pip/rapids_dask_dependency/pyproject.toml
vyasr: Yes I think so.

vyasr:
Thanks for all the feedback here! As per discussions both here and offline, we're moving this metapackage into its own repository. All the PR feedback has been addressed AFAIK and future work can continue on https://github.com/rapidsai/rapids-dask-dependency.

PR #690, opened 2023-11-18
Add script to test nightly environments are solvable and using recent nightlies.

bdice:
This PR adds tests that run on every `integration` PR and on a nightly basis to ensure that RAPIDS conda environments can be solved with recent packages. This will help us diagnose and react to problems arising from conda environment conflicts, which sometimes force conda to select older nightly builds of RAPIDS packages.

A partial solution for https://github.com/rapidsai/ops/issues/2947.

Next steps:
- [x] Wait for https://github.com/rapidsai/cugraph/pull/4639 to merge.
- [x] After this PR is merged, merge https://github.com/rapidsai/workflows/pull/59 as well.


ci/test_conda_nightly_env.sh
```diff
@@ -0,0 +1,29 @@
...
+
+#rapids-conda-retry \
+conda \
+    create \
+    --solver=libmamba \
+    -n rapids-${RAPIDS_VERSION} \
+    -c rapidsai-nightly \
+    -c conda-forge \
+    -c nvidia  \
+    rapids=${RAPIDS_VERSION} \
```
jameslamb, line 22: *(was going to put this as my review, realized a thread would probably be better)*

I totally support adding something like this, based on your description offline:

> _... to ensure that we are able to solve the full RAPIDS conda environment with recent packages (in other words, ensure there are no recent conflicts causing fallback to older conda packages)_

But I don't think it offers 100% protection against the case described in https://github.com/rapidsai/ops/issues/2947.

That issue is not about just packages building against too-old versions of dependencies... it's about packages across RAPIDS building against **very different**  versions of dependencies.

It looks to me that the code in this PR would catch cases like these:

- *"`rmm` nightlies haven't been published in the last 3 days"*
- *"the latest versions of `cugraph` and `pylibraft` can't be installed in the same environment"*

These aren't captured by the existing nightly tests at https://github.com/rapidsai/workflows/actions/workflows/nightly-pipeline.yaml.

But this wouldn't be guaranteed to catch a case like this:

- *"the latest `cuml` nightly built against an `rmm` from 5 days ago, but the latest `cudf` nightly built against an `rmm` from yesterday"*

Because this test with the `rapids` package is solving across all of the packages' **runtime** dependencies, but they could have ended up building against older versions based on conflicts in their individual **build** environments, right?

And those types of conflicts might not show up here if we use `pin_compatible(max_pin="x.x")` in `run` dependencies, e.g. `pylibraft==24.10.*` is going to have a runtime dependency on `rmm=24.10.*` regardless of which specific nightly of `rmm` it pulled in at build time. ([pylibraft nightly files](https://anaconda.org/rapidsai-nightly/pylibraft/files))

I think detecting that other case would have to happen at build time (or by post-processing of logs from build time). And I don't know how complex that would be, so can't say with confidence that the complexity would be worth it.

I totally support the approach this PR is pursuing, just wanted to be sure to note this other possible avenue for version mismatches to get through.

jameslamb (commented):
I totally support this!

Think it'll be helpful for catching more cases similar to https://github.com/rapidsai/build-planning/issues/14 and https://github.com/rapidsai/build-planning/issues/69 where projects might otherwise silently be using different packages than we want.

ci/test_conda_nightly_env.sh
bdice: You are correct about what this will and will not cover. I think it's worth pursuing because (1) it prevents runtime problems from being hidden and (2) sometimes runtime conflicts will also affect build environments, so this may give us a bit of signal into deeper problems should they arise.

jameslamb: Ok great! I totally support moving forward with this.

bdice:
Yay! I'm happy with this now. It looks like this:
image

.github/workflows/pr.yaml
```diff
@@ -10,13 +10,21 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  pr-builder:
+    needs:
+      #- build
```
jameslamb, line 15:
```suggestion
      - build
```

I'm guessing this is just leftover from testing, and should this check should be enforced again.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,23 @@
...
+    inputs:
+      branch:
+        required: true
+        type: string
+      date:
+        required: true
+        type: string
+      sha:
+        required: true
+        type: string
```
jameslamb, line 14:
```suggestion
```

It looks to me like these inputs aren't being used for anything. If that's intentional, they should be removed.

jameslamb (approved):
This looks great!

Although I'll note... I wasn't able to see the pretty colored output that you shared a screenshot of. 

The 13,000+ lines of logs at
https://github.com/rapidsai/integration/actions/runs/10622584645/job/29447212735?pr=690 was too large for my browser to load. The majority of that was what looks like a dump of every conda package in the environment's metadata, in pretty-printed (one-line-per-key) JSON format

image

I was able to open the raw logs ([link](https://productionresultssa0.blob.core.windows.net/actions-results/66341d0a-4a04-40f4-8dff-ff35453933e9/workflow-job-run-3ee3e4ba-dc6f-540a-3290-7fe7e8fdc66e/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-08-29T22%3A40%3A48Z&sig=x8DampCLTDVF0NyRPNlphW02S65p%2Bn3vAagXkGTQL9I%3D&ske=2024-08-30T07%3A19%3A12Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2024-08-29T19%3A19%3A12Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2024-05-04&sp=r&spr=https&sr=b&st=2024-08-29T22%3A30%3A43Z&sv=2024-05-04)) and command-F and find the output from this job, but that's not the same

image

If you can think of a way to reduce the amount of output or to get just the color-coded results up into the job summary, I think it'd help.

.github/workflows/test.yaml
bdice: These are used by https://github.com/rapidsai/workflows/pull/59. It's mostly just to align with what every other RAPIDS repo does to track what dates/SHAs are tested.

.github/workflows/pr.yaml
bdice: Fixed!

bdice:
We can merge this once CI passes. I'm going to check the CI results manually so I'm avoiding a `/merge` for the moment.

bdice:
This is failing due to a "real" problem now, I think.
image

Possibly related to https://github.com/rapidsai/cuspatial/pull/1453#issuecomment-2334885379.

PR #701, opened 2024-02-21
Adds nx-cugraph to the rapids meta-package

rlratzel:
* Adds `nx-cugraph` to the `rapids` meta-package - unlike Accelerated Pandas, which is provided with `cudf`, Accelerated NetworkX is not provided with `cugraph` and must be installed separately.
* `nx-cugraph` has a dependency on NetworkX and ensures a compatible version is installed.  Because of this, the explicit install of `networkx` is removed from the `rapids` meta-package in favor of using the `networkx` version installed via `nx-cugraph`.


rlratzel:
cc @beckernick

raydouglass (approved):

PR #712, opened 2024-06-28
Increase maximum UCX runtime pin to `<1.18.0`

pentschev:
Increase maximum UCX runtime pin to support UCX up to `v1.17.x`, but support `v1.15.0` and above until we gain confidence with newer UCX.


msarahan (approved):

jameslamb (approved):
Giving you a `packaging-codeowners` approval, this looks ok to me.

raydouglass (approved):

PR #715, opened 2024-08-02
Remove numba dependency

gmarkall:
**EDIT:** This now removes the numba dependency, which comes transitively from cuDF.

**Original description:**

numba-cuda is the NVIDIA-maintained CUDA target for Numba, which depends on the numba package.

This PR replaces the numba dependency with numba-cuda. However, similar to @bdice's comment in #648, I don't know if we strictly need this here any more due to the cudf dependency (from rapidsai/cudf#16474) ensuring numba-cuda is present.


bdice:
I think we can delete the dependencies that come transitively from RAPIDS repositories. Let’s remove the numba pins here.

conda/recipes/rapids/meta.yaml
```diff
@@ -31,7 +31,7 @@ requirements:
     - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
     - cupy {{ cupy_version }}
     - nccl {{ nccl_version }}
-    - numba {{ numba_version }}
+    - numba-cuda {{ numba_cuda_version }}
```
gmarkall, line 34:
```suggestion
```

conda/recipes/versions.yaml
```diff
@@ -8,8 +8,8 @@ cupy_version:
   - '>=12.0.0'
 nccl_version:
   - '>=2.9.9,<3.0a0'
-numba_version:
-  - '>=0.57'
+numba_cuda_version:
+  - '>=0.0.13'
```
gmarkall, line 12:
```suggestion
```

gmarkall:
@bdice Many thanks for the suggestion - I think I've implemented it as you suggested (and updated the title / description)

bdice (approved):

bdice:
@gmarkall I think this might be ready to merge if you agree. I merged the upstream since CI was failing.

gmarkall:
@bdice I think this should be fine as the dependency comes from cudf (and will come from its numba-cuda dependency once https://github.com/rapidsai/cudf/pull/16474 is merged).

bdice:
I updated this PR and retargeted to 24.12. I will give it a `/merge` so it can move forward if CI passes.

PR #717, opened 2024-08-07
Improve update-version.sh

bdice:
A few small tweaks to `update-version.sh` for alignment across RAPIDS.


jameslamb (approved):
nice! Glad we could standardize on `sed_runner` to get portability.

PR #718, opened 2024-08-19
Remove NumPy <2 pin

seberg:
This PR removes the NumPy<2 pin which is expected to work for
RAPIDS projects once CuPy 13.3.0 is released (CuPy 13.2.0 had
some issues preventing the use with NumPy 2).

PR #720, opened 2024-09-18
DO NOT MERGE: force numpy >=2

msarahan:
This is for testing purposes. Once this PR is passing, it is safe to merge https://github.com/rapidsai/integration/pull/718, which allows numpy range from >=1.23,<3.0a0


msarahan:
James suggested that it would be simpler to just use #718 for this purpose. Closing.

PR #752, opened 2025-02-27
Replace `conda-mambabuild` with `conda-build`

jakirkham:
These are now equivalent in behavior. However support for `conda-mambabuild` is being dropped. So switch to `conda-build`.

xref: https://github.com/rapidsai/build-planning/issues/149

### DIFF
--- a/ci/build_python.sh
+++ b/ci/build_python.sh
@@ -15,14 +15,14 @@ rapids-print-env
 
 rapids-logger "Build rapids-xgboost"
 
-rapids-conda-retry mambabuild \
+rapids-conda-retry build \
   --use-local \
   --variant-config-files "${CONDA_CONFIG_FILE}" \
   conda/recipes/rapids-xgboost
 
 rapids-logger "Build rapids"
 
-rapids-conda-retry mambabuild \
+rapids-conda-retry build \
   --use-local \
   --variant-config-files "${CONDA_CONFIG_FILE}" \
   conda/recipes/rapids
--- a/ci/build_python.sh
+++ b/ci/build_python.sh
@@ -16,14 +16,12 @@ rapids-print-env
 rapids-logger "Build rapids-xgboost"
 
 rapids-conda-retry build \
-  --use-local \
   --variant-config-files "${CONDA_CONFIG_FILE}" \
   conda/recipes/rapids-xgboost
 
 rapids-logger "Build rapids"
 
 rapids-conda-retry build \
-  --use-local \
   --variant-config-files "${CONDA_CONFIG_FILE}" \
   conda/recipes/rapids
 


ci/build_python.sh
```diff
@@ -16,14 +16,12 @@ rapids-print-env
 rapids-logger "Build rapids-xgboost"
 
 rapids-conda-retry build \
-  --use-local \
```
jakirkham, line 19: As `conda-build` always uses its local cache, the `--use-local` flag is unneeded

Additionally it appears `conda-build` has a bug that causes it to look inside its package cache before it is created, which causes [the following CI error]( https://github.com/rapidsai/integration/actions/runs/13558475509/job/37897248372?pr=752#step:9:413 )

```
UnavailableInvalidChannel: HTTP 404 UNAVAILABLE OR INVALID for channel opt/conda/conda-bld <file:///opt/conda/conda-bld>

The channel is not accessible or is invalid.

You will need to adjust your conda configuration to proceed.
Use `conda config --show channels` to view your configuration's current state,
and use `conda config --show-sources` to view config file locations.

As of conda 4.3, a valid channel must contain a `noarch/repodata.json` and
associated `noarch/repodata.json.bz2` file, even if `noarch/repodata.json` is
empty. Use `conda index /opt/conda/conda-bld`, or create `noarch/repodata.json`
and associated `noarch/repodata.json.bz2`.
```

Typically we don't use `/opt/conda/conda-bld`, but use [our own package cache location]( https://github.com/rapidsai/ci-imgs/blob/5a340212e24926cdac1e895f7d790418070bce2c/context/condarc.tmpl#L12 ). So this isn't even the right place for it to check

Since we don't need this flag here (nor use it elsewhere), it is easy enough to drop, which fixes all of these issues

jakirkham:
The `test-conda-nightly` env failures are unrelated and appear to be due to issue: https://github.com/rapidsai/rapids-logger/issues/27

jameslamb (commented):
Even if that `rapids-logger` change was made, you'd still run into issues like what you're facing here because that project isn't built every night (https://github.com/rapidsai/rapids-logger/actions/workflows/build.yaml).

It'd be better to just add `rapids-logger` to this exclusion list:

https://github.com/rapidsai/integration/blob/3b43bff142c7f75ab88d0d970a779ccf400b5d51/ci/check_conda_nightly_env.py#L10-L13

ci/build_python.sh
jameslamb: Works for me. I looked through the blame to check why it was added originally... looks like you added it, I think to be sure that the build of `integration` uses the `rapids-xgboost` just built in the same build:

https://github.com/rapidsai/integration/pull/669#discussion_r1275517623

So I checked CI logs here to be sure this was working as expected and not falling back to a previous `rapids-xgboost` nightly.

It is:

```text
BUILD START: ['rapids-xgboost-25.04.00a-cuda11_py310_250227_g07bf842_11.conda']
...
BUILD START: ['rapids-25.04.00a-cuda11_py310_250227_g07bf842_11.conda']
...
    rapids-xgboost:                   25.04.00a-cuda11_py310_250227_g07bf842_11        file:///tmp/conda-bld-output
...
```

([build link](https://github.com/rapidsai/integration/actions/runs/13574946373/job/37948869163?pr=752#step:9:1014))

jakirkham: Good catch!

Maybe there was a past conda-build issue we were working around with the test phase

jakirkham:
Thanks James! 🙏

PR #756, opened 2025-04-08
Compute matrix for nightly conda environment checks.

bdice:
Currently CI is failing because we are using `conda-python-build.yaml` for a CPU test job. Some changes were made in https://github.com/rapidsai/shared-workflows/pull/297 that make this workflow less suitable for this generalized usage. Instead, we should compute a custom matrix and use that.

### DIFF
--- a/.github/actions/compute-matrix/action.yaml
+++ b/.github/actions/compute-matrix/action.yaml
@@ -1,0 +1,43 @@
+name: Compute matrix
+
+outputs:
+  MATRIX:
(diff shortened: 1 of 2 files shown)


AyodeAwe:
There might be a syntax error in the `pr.yaml` workflow

bdice:
@AyodeAwe Hmm. I will need to dig into this again to figure out the problem but I have a lot on my plate at the moment. If you want to push to this PR, please do!

PR #772, opened 2025-05-27
Port all conda recipes to rattler-build

gforsyth:
xref rapidsai/build-planning#47

Supercedes #771 which is a history mess

### DIFF
--- a/conda/recipes/rapids-xgboost/recipe.yaml
+++ b/conda/recipes/rapids-xgboost/recipe.yaml
@@ -31,13 +31,9 @@ requirements:
     - python
     - libxgboost ${{ xgboost_version }} rapidsai_h*
     - xgboost ${{ xgboost_version }} rapidsai_py*
-    - if: cuda_major == "11"
-      then: cudatoolkit
   ignore_run_exports:
     by_name:
       - python_abi
-      - if: cuda_major == "11"
-        then: cudatoolkit
 
 about:
   homepage: https://rapids.ai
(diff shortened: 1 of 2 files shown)


conda/recipes/rapids-xgboost/recipe.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - python
+    - libxgboost ${{ xgboost_version }} rapidsai_h*
+    - xgboost ${{ xgboost_version }} rapidsai_py*
+    - if: cuda_major == "11"
+      then: cudatoolkit
+  ignore_run_exports:
+    by_name:
+      - python_abi
+      - if: cuda_major == "11"
+        then: cudatoolkit
```
jameslamb, line 40:
```suggestion
```

Now that CUDA 11 support is getting dropped on 25.08 (https://github.com/rapidsai/build-planning/issues/184), I think we could safely drop all the `if: cuda_major == "11"` stuff in any files you're already touching in this PR.

Notice that there aren't any CUDA 11 builds at https://github.com/rapidsai/integration/actions/runs/15421036533/job/43395979955?pr=772

image

ci/build_python.sh
```diff
@@ -3,24 +3,26 @@
 
 set -euo pipefail
 
-rapids-configure-conda-channels
```
jameslamb, line 6: 🎉 thank you! Getting closer on https://github.com/rapidsai/gha-tools/issues/145 😁

jameslamb (approved):
Looks great to me!

This https://github.com/rapidsai/integration/blob/branch-25.08/conda/recipes/versions.yaml is one bit that's a bit different here compared to other RAPIDS projects, so I double-checked build logs to be sure those versions were getting pulled through... all looks good to me.

I left one note about removing CUDA 11 stuff in this PR. Do what you want with it, I don't feel strongly about doing it now vs. later in a follow-up PR.

Issue #773, opened 2025-05-28
25.08 nightlies failing: cannot solve the test environment for `rapids`


jameslamb:
## Description

`branch-25.08` nightly builds are all failing.

CUDA 11 jobs appear to get stuck solving the test environment for `rapids` then hit the 6-hour timeout enforced by GitHub Actions:

```text
Preparing transaction: ...working... done
Verifying transaction: ...working... done
Executing transaction: ...working... done
Reloading output folder: ...working... done
context canceled
Solving environment (_test_env): ...working... 
```

([build link](https://github.com/rapidsai/integration/actions/runs/15300765008/job/43040523188))

CUDA 12 jobs fail outright after about an hour, also with solver errors setting up the test environment for `rapids`:

```text
Solving environment (_test_env): ...working... failed
WARNING: failed to get package records, retrying.  exception was: Unsatisfiable dependencies for platform linux-64: {MatchSpec("cuda-version[version='>=11,<12.0a0']"), MatchSpec("cucim==25.08.00a7=cuda11_py310_250521_g66bbd8a_7")}
Encountered problems while solving:
  - package cucim-25.08.00a7-cuda11_py310_250521_g66bbd8a_7 requires cuda-version >=11,<12.0a0, but none of the providers can be installed
```

([build link](https://github.com/rapidsai/integration/actions/runs/15300765008/job/43040524735#step:10:5270))

<details><summary>full solver traceback (click me)</summary>

The full solver traceback exceeds GitHub's limits for issue bodies 🙃 

## Reproducible Example

Any recent PR or nightly builds of `branch-25.08`.

No problems on `branch-25.06`.

## Notes

N/A

jakirkham:
AIUI we are missing XGBoost packages for 25.08. This is being fixed with the builds of PR: https://github.com/rapidsai/xgboost-feedstock/pull/93

Should be able to restart after that

jakirkham:
The RAPIDS XGBoost 25.08 nightly packages are now up! 🎉

Also a recent 25.08 forward merger ( https://github.com/rapidsai/integration/pull/776 ) updating the XGBoost version constraint, just started fresh builds for 25.08

Let's see how those go

jakirkham:
So the CUDA 12 builds pass

However the CUDA 11 builds now fail. The reason being RAPIDS 25.08 has XGBoost 3.0.2, which is only compatible with CUDA 12

[...]

PR #774, opened 2025-05-28
WIP: investigate failed 25.08 solves

jameslamb:
Contributes to #773 

Just opening this to test some theories for failing 25.08 nightlies.


jameslamb:
Turned out that the failures were missing `xgboost` packages, and that this is resolved by changes over in https://github.com/rapidsai/xgboost-feedstock/pull/93

We don't need this testing PR any more.

PR #757, opened 2025-04-18
Remove cuspatial and cuproj from rapids metapackage.

bdice:
`cuspatial` and `cuproj` can be dropped from the `rapids` metapackage.

See https://docs.rapids.ai/notices/rsn0045/.

PR #781, opened 2025-07-15
WIP: remove cuspatial and cuproj from rapids metapackage

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/197

Removes `cuspatial` and `cuproj` from the runtime dependencies of the `rapids` metapackage. The final releases of those packages were in RAPIDS 25.04.

### Didn't we already do this?

Yes, #757 removed this for 25.06 (and that was forward-merged to 25.08).

But it looks like they were accidentally re-introduced in 25.08 via #772

And that technically worked because a small number of `cuspatial` and `cuproj` 25.08 nightlies did make it onto the `rapidsai-nightly` channel before we fully stopped development of `cuspatial`.


gforsyth (approved):
:shipit:

PR #786, opened 2025-08-01
Update xgboost version to 3.0.3

AyodeAwe:
Update xgboost to match the feedstock for `25.08`


jakirkham (commented):
Thanks Jake! 🙏

I had a typo in my previous suggestion. Have added a new suggestion below to fix this

PR #788, opened 2025-08-03
Add correct rapids version environment variable

AyodeAwe:
Adds changes probably missed in https://github.com/rapidsai/integration/pull/772

cc @gforsyth


AyodeAwe:
The failing tests are probably expected due to 25.10 being never published yet.

PR #792, opened 2025-08-04
Set pre-commit.ci config to only update quarterly

bdice:
This adds config for pre-commit.ci to file fewer automatic PRs.

Closes #791.


jameslamb (approved):
Support it!

PR #770, opened 2025-05-27
stop uploading to downloads.rapids.ai, add shellcheck

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/181

* removes all uploads of conda packages and wheels to `downloads.rapids.ai`

Contributes to https://github.com/rapidsai/build-planning/issues/135

* adds `shellcheck` checks to `pre-commit` configuration

~Also proposes changes to get around the cold-start problem we have at the beginning of every release, where `test-conda-nightly-env` cannot run because there aren't yet published `rapids` packages.~

* ~makes `test-conda-nightly-env` dependent on the just-built-in-CI `rapids` packages, instead of installing them from `rapidsai-nightly`~

### How I identified changes

Looked for uses of the relevant `gha-tools` tools, as well as documentation about `downloads.rapids.ai`, being on the NVIDIA VPN, using S3, etc. like this:

```shell
git grep -i -E 's3|upload|downloads\.rapids|vpn'
```

### How I tested this

See "How I tested this" on https://github.com/rapidsai/shared-workflows/pull/364

PR #795, opened 2025-08-26
Build and test with CUDA 13.0.0

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/208

* uses CUDA 13.0.0 to build and test
* moves some dependency pins:
  - `cupy`: `>=13.6.0`

## Notes for Reviewers

This switches GitHub Actions workflows to the `cuda13.0` branch from here: https://github.com/rapidsai/shared-workflows/pull/413

A future round of PRs will revert that back to `branch-25.10`, once all of RAPIDS supports CUDA 13.

### What about XGBoost?

This proposes the following TEMPORARILY for CUDA 13:

* not building `rapids-xgboost` packages
* `rapids` metapackage depending on CPU builds of `xgboost` from conda-forge

This will allow us to start publishing CUDA 13 nightlies of the `rapids` metapackage and other things that depend on it, like the RAPIDS container images (https://github.com/rapidsai/docker/pull/782).

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -34,7 +34,7 @@ concurrency:
 jobs:
   build:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@branch-25.10
+    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@cuda13.0
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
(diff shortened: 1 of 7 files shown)


jameslamb:
The test failures like this are expected:

>   "error": "LibMambaUnsatisfiableError: Encountered problems while solving:\n  - package rapids-25.10.00a6-cuda12_py310_250828_8ca32656 requires cuda-version >=12,<13.0a0, but none of the providers can be installed\n\nCould not solve for environment specs\nThe following packages are incompatible\n\u251c\u2500 cuda-version =13.0 * is requested and can be installed;\n\u2514\u2500 rapids =25.10 * is not installable because it requires\n   \u2514\u2500 cuda-version >=12,<13.0a0 *, which conflicts with any installable versions previously reported.",

([test-conda-nightly env link](https://github.com/rapidsai/integration/actions/runs/17476585420/job/49637942518?pr=795#step:10:29)

Those tests only work on already-published packages, so there's always a need for an admin merge during migrations like this.

That's intentional: https://github.com/rapidsai/integration/pull/770#issuecomment-2922588436

In other words... once we admin-merge this, that test should start passing in CI on future PRs.

Issue #794, opened 2025-08-19
Test Deep Learning Frameworks in nightlies


bdice:
@caryr35 and I were discussing and we would like to add automated nightly tests that spot-check RAPIDS + Deep Learning Frameworks.

Currently there is a manual process for running smoke tests. It is mostly imports and ensuring GPU support is available.

We can automate this process with CI for our nightlies.

### PyTorch
First, an environment is created:
```bash
conda create -n rapids-25.08 -c rapidsai -c conda-forge -c nvidia rapids=25.08 python=3.13 'cuda-version>=12.0,<=12.9' 'pytorch=*=*cuda*'
```

Then the following smoke tests are run:
```python
import cudf
import cuml
import cugraph
import torch
assert torch.cuda.is_available()
```

### TensorFlow
First, an environment is created:
(Note: TensorFlow doesn't support Python 3.13 at the time of writing)
```bash
conda create -n rapids-25.08 -c rapidsai -c conda-forge -c nvidia rapids=25.08 python=3.12 'cuda-version>=12.0,<=12.9' tensorflow
```

Then the following smoke tests are run (modified slightly to assert rather than print):
```python
import cudf
import cuml
import cugraph
import tensorflow as tf
gpus = tf.config.list_physical_devices("GPU")
print(gpus)
assert len(gpus) > 1
```

PR #837, opened 2026-04-07
feat: add stable install tests

### LINKED ISSUE #120: [FEA] Nightly build of latest stable
**Is your feature request related to a problem? Please describe.**

Issue https://github.com/rapidsai/cudf/issues/6096 caused an expensive fire drill involving core members across multiple projects

Assuming the current stable release breaks again in the future for one of many reasons, and stable releases are out for ~6w, a useful thing may be something along the lines of daily build tests of last 2 releases (3mo)

**Describe the solution you'd like**

* ~Daily builds of the last ~2 stable releases, giving coverage of releases for last ~3mo
* Tests standalone (per-repo) + combined
* For case of individual repos fine but combined failing, treat as a cudf issue (?)
* Visibility - regular user: Version status badge ("0.14 passing") on each repo (self + global)
* Visibility - new user: rapids.ai site
* Visibility - core team: ?

I'm less clear on the relative value of the full gamut of going across CUDA etc versions. AFAICT, the 80% is probably around conda: Python verison x set of RAPIDS packages

gforsyth:
This PR adds a nightly test that attempts to install the latest stable version
of RAPIDS on all supported Python and CUDA versions, on amd64 and arm64, with
both `pip` and `conda`.

The tests for this pass consist of:

- does it install successfully (and for pip, can we install the available packages from upstream pypi)
- can we import the installed libraries:
  - individually: this checks whether each library is explicitly loading all the symbols it requires
  - concurrently: this checks for things like libraries linking against different (and incompatible) versions of dependencies

Import success is defined as:
- Python exits with a clean exit code after running `import foo`
- Stdout and stderr from the import do not contain text indicating missing symbols, failed initialization, etc
  - the array of strings we consider failure markers is in `ci/stable_install/test_imports.sh` and is used for both `conda` and `pip` imports

Initial implementation of https://github.com/rapidsai/build-planning/issues/227.


ci/stable_install/install_and_test_conda.sh
```diff
@@ -0,0 +1,68 @@
...
+
+        rapids-logger "Testing stable version install with Python $PY_VER and CUDA version $CUDA_VERSION"
+
+        # use `-O` to override channels so we don't include `rapidsai-nightly`
+        conda create -n "$envName" -O -c rapidsai -c conda-forge \
+          rapids="$STABLE_RAPIDS_VERSION" python="$PY_VER" "cuda-version>=${CUDA_VERSION}" -y
+
+        conda activate "$envName"
+
+        testImports cudf dask_cudf cuml pylibraft raft_dask cugraph nx_cugraph cuxfilter cucim cuvs
```
jameslamb, line 58:
```suggestion
       python -c "import cudf; import dask_cudf; import cuml; import pylibraft; import raft_dask; import cugraph; import nx_cugraph; import cuxfilter; import cucim; import cuvs;"
```

Up to you on spelling (maybe something with concatenating elements of a bash array would be easier to modify) but I think this would be a more powerful test as a single import of everything all at once.

That gives us a chance to catch problems like "cuML and cuGraph linked against different, incompatible NCCL versions" or similar.

ci/stable_install/install_and_test_conda.sh
```diff
@@ -0,0 +1,68 @@
...
+
+for CUDA_VERSION in "${SUPPORTED_CUDA_VERSIONS[@]}"; do
+    for PY_VER in "${SUPPORTED_PYTHON_VERSIONS[@]}"; do
+        envName="rapids_${PY_VER}_${CUDA_VERSION}"
+
+        rapids-logger "Testing stable version install with Python $PY_VER and CUDA version $CUDA_VERSION"
+
+        # use `-O` to override channels so we don't include `rapidsai-nightly`
+        conda create -n "$envName" -O -c rapidsai -c conda-forge \
+          rapids="$STABLE_RAPIDS_VERSION" python="$PY_VER" "cuda-version>=${CUDA_VERSION}" -y
```
jameslamb:
```suggestion
        conda create -n "$envName" -O -c rapidsai -c conda-forge -y \
          rapids="$STABLE_RAPIDS_VERSION" python="$PY_VER" "cuda-version>=${CUDA_VERSION}"
```

Very very minor style suggestion... the `-y` was hard to see at the end of the set of requirements. It'd be clearer grouped with the other arguments, I think.

ci/stable_install/install_and_test_conda.sh
```diff
@@ -0,0 +1,68 @@
...
+# By default, this script attempts to install the latest stable RAPIDS version
+# in a clean conda environment for all supported versions of Python and CUDA
+#
+# The Python version, CUDA version, and RAPIDS version can be overridden by
+# supplying the value to test to the appropriate flag (--python, --rapids, and
+# --cuda, respectively).
+
+set -euo pipefail
+
+STABLE_RAPIDS_VERSION="26.4.*"
```
jameslamb, line 16: https://github.com/rapidsai/integration/blob/main/ci/release/update-version.sh should be updated to modify this

ci/stable_install/install_and_test_conda.sh
```diff
@@ -0,0 +1,68 @@
...
+SUPPORTED_PYTHON_VERSIONS=(3.11 3.12 3.13 3.14)
+SUPPORTED_CUDA_VERSIONS=("12.9" "13.1")
```
jameslamb:
```suggestion
SUPPORTED_CUDA_VERSIONS=("12.2" "12.9" "13.0" "13.1")
```

What do you think about expanding this to the upper and lower bounds RAPIDS supports? The older versions tend to be tested less thoroughly in CI and developer environments, so I could see some types of integration issues making it through undetected.

ci/stable_install/install_and_test_pip.sh
```diff
@@ -0,0 +1,97 @@
...
+# supplying the value to test to the appropriate flag (--python, --rapids, and
+# --cuda, respectively).
+#
+# Note that the CUDA version should be specified as a -cu* suffix, e.g. `--cuda cu13`
+
+set -euo pipefail
+
+STABLE_RAPIDS_VERSION="26.4.*"
+SUPPORTED_PYTHON_VERSIONS=(3.11 3.12 3.13 3.14)
+SUPPORTED_CUDA_VERSIONS=("cu12" "cu13")
```
jameslamb, line 18: Would you consider also having this test the bounds (12.2, 12.9, 13.0, 13.1)?

That could be done by adding a requirement like this to the `pip install` calls:

```text
cuda-toolkit[all]==${cuda_major_minor}.*
```

That'd be a really helpful extension of https://github.com/rapidsai/build-planning/issues/256, and I think it'd help us catch conflicts that aren't easily caught in individual projects' CI.

ci/stable_install/install_and_test_pip.sh
```diff
@@ -0,0 +1,97 @@
...
+
+    PIP_INSTALL_NVIDIA_PYPI=("${PIP_INSTALL_PYPI[@]}"
+            "cugraph-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}"
+            "nx-cugraph-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}"
+            "cuxfilter-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}"
+            "cucim-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}"
+            "cuvs-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}"
+            --extra-index-url=https://pypi.nvidia.com
+            )
+
```
jameslamb: Similar to my comments on conda, I think this would be a more powerful test if it combined all the imports into one.

I'd structure it roughly like this:

```shell
# get all the pypi.nvidia.com stuff
wheels_dir=$(mktemp -d)
pip download \
  --isolated \
  --index-url https://pypi.nvidia.com \
  --prefer-binary \
  --no-deps \
  -d "${wheels_dir}" \
  "cucim-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}" \
  "cugraph-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}" \
  "cuvs-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}" \
  "cuxfilter-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}" \
  "libcugraph-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}" \
  "libcuvs-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}" \
  "pylibcugraph-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}"

pip install \
  --isolated \
  --index-url https://pypi.org/simple \
  --prefer-binary \
  "${PIP_INSTALL_PYPI[@]}" \
  "${wheels_dir}"/*.whl

python -c "import cudf; import dask_cudf; ...; import cuvs"
```

Would you consider something like that?

.github/workflows/test.yaml
```diff
@@ -28,3 +28,63 @@ jobs:
       script: "ci/test_conda_nightly_env.sh"
       # just using the workflow to get the matrix, this isn't actually building anything we want to upload
       upload-artifacts: false
+  test-stable-install-pip-cuda-12-amd64:
+    secrets: inherit
```
jameslamb:
```suggestion
```

none of these jobs need any secrets, as far as I can tell. Can you please remove these, to limit what this workflow's code has access to?

.github/workflows/test.yaml
```diff
@@ -28,3 +28,63 @@ jobs:
...
+    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@main
+    with:
+      build_type: pull-request
+      node_type: "cpu4"
+      arch: "amd64"
+      container_image: "nvidia/cuda:12.9.1-base-ubuntu-24.04"
+      script: |
+        ./ci/stable_install/install_and_test_pip.sh --cuda cu12
+  test-stable-install-pip-cuda-13-amd64:
```
jameslamb: We could cut the number of jobs here in half by having `matrix` elements inside of each of these for amd64 and arm64. Like this: https://github.com/rapidsai/cuvs/blob/cbb9db5697eeebcb03a6ed198b7d4386ce14a301/.github/workflows/pr.yaml#L365-L374

Will you consider that?

jameslamb (commented):
Thanks for doing this!

I nosed in and left a couple comments for your consideration... I've spent the last couple weeks deep in RAPIDS' wheel-testing setup (e.g. for https://github.com/rapidsai/build-planning/issues/256), and that informed some of what I was looking for here.

Overall I do really like the structure! Especially running the jobs in parallel and using the `nvidia/cuda:*-base-*` image for wheels.

And I agree that "install everything + import the libraries" is a good target for this first pass.

ci/stable_install/install_and_test_conda.sh
gforsyth: interesting -- I think I'll add this as an additional import test.

My original intent was to catch libraries that don't load all the necessary shared objects, and importing everything at once might let something slip through, but checking for incompatible linking is a good check, too!

.github/workflows/test.yaml
jameslamb: That language was a little imprecise... cut the number of **configurations** in half. The number of jobs would be unchanged.

ci/stable_install/install_and_test_conda.sh
jameslamb: Ok yeah, even better!

ci/stable_install/install_and_test_conda.sh
gforsyth: Done!

ci/stable_install/test_imports.sh
```diff
@@ -0,0 +1,21 @@
...
+        rapids-logger "Passed"
+        # add import to array for combined import test before shifting
+        imports+=("$1")
+        shift
+    done
+    import_cmd=$(printf "import %s; " "${imports[@]}")
+    rapids-logger "Combined import test for: ${imports[*]}"
+    python -c "${import_cmd}" || rapids-logger "Test failed for: ${imports[*]}"
+    rapids-logger "Passed"
+}
```
gforsyth, line 31: This does enough now that I broke it out into a separate script so it can be `source`d by both test scripts.

Imports each library individually in separate Python sessions, then imports all of them sequentially in the same Python session.

ci/stable_install/install_and_test_pip.sh
```diff
@@ -0,0 +1,114 @@
...
+            --no-deps \
+            -d "${WHEELS_DIR}" \
+            "cucim-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}" \
+            "cugraph-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}" \
+            "cuvs-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}" \
+            "cuxfilter-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}" \
+            "libcugraph-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}" \
+            "libcuvs-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}" \
+            "nx-cugraph-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}" \
+            "pylibcugraph-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}"
```
gforsyth, line 103: This does an unnecessary double-download for each CUDA major version, but avoiding it requires annoying and error-prone state-tracking

ci/stable_install/test_imports.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+# Copyright (c) 2026, NVIDIA CORPORATION.
+
+set -euo pipefail
+
+function testImports {
+    unset imports
```
jameslamb:
```suggestion
    local -a imports=()
```

Would something like this be slightly safer than `unset`? I think (untested) this would create a function-scoped `imports`  on each call.

ci/stable_install/test_imports.sh
jameslamb: I like it a lot!!!

ci/stable_install/test_imports.sh
```diff
@@ -0,0 +1,21 @@
...
+    while [[ $# -gt 0 ]]; do
+        # run standalone import test
+        rapids-logger "Standalone import test for $1"
+        python -c "import $1" || rapids-logger "Test failed for: $1"
```
jameslamb: I don't think this will fail the CI jobs when the imports fail, and I think it should, right? I think we want a big ❌ and a job failure, not for us to need to remember to go read the logs.

This `||` is going to swallow the exit code of `python -c "import $1"` and the pipe will exit with `0` because that logging statement will succeed.

And I don't see any other logic that's doing something like "grep for `Test failed` and exit 1 if you find any" in the calling scripts.

I recommend hard-coding in like `python -c "import some_nonsense"` here and checking that the CI job fails when we want.

ci/stable_install/install_and_test_pip.sh
jameslamb: Thanks for calling it out. I think that's totally fine.

jameslamb (changes requested):
Thanks for the import changes, I think we'll be really happy to have those!

I left a couple more suggestions, and one that I think is worth blocking the PR over.

I'd love to also see actual runs of these jobs. Since this was opened from your fork, we won't be able to trigger `test.yaml` manually with workflow dispatch... could you temporarily add these jobs to `pr.yaml` so we can see them in PR CI? That could be reverted before this is merged.

ci/stable_install/test_imports.sh
gforsyth: Yep, you're right.  For now I'm going to remove the `||` and let the errors percolate up.  I can put up a follow-up at some point that sticks all the output in a named pipe or something and then greps over all the output for errors

gforsyth:
Ahh, ok, so `shared-workflows` definitely assumes we're running a RAPIDS image, because `rapids-github-info` requires that `git` is installed, and the `sccache` setup step assumed that `curl` and `jq` are installed.

Even the beefy 5 GB+ cuda-devel images don't have those dependencies installed.  

I think I'm going to open a PR to add a "bootstrap script" option to `custom_job`, that can execute quick install commands before the rest of the `shared-workflow` steps get kicked off.

gforsyth:
> Ahh, ok, so `shared-workflows` definitely assumes we're running a RAPIDS image, because `rapids-github-info` requires that `git` is installed, and the `sccache` setup step assumed that `curl` and `jq` are installed.
> 
> Even the beefy 5 GB+ cuda-devel images don't have those dependencies installed.
> 
> I think I'm going to open a PR to add a "bootstrap script" option to `custom_job`, that can execute quick install commands before the rest of the `shared-workflow` steps get kicked off.

Even with the bootstrap script this doesn't work.
`actions/checkout` runs before any script is executed and because the `nvidia/cuda` images don't have `git` installed, `checkout` pulls a tarball snapshot instead, so then we don't have a proper repo cloned down.

I genuinely didn't think that creating our own version of `nvidia/cuda` that also includes `curl`, `git`, and `jq` would be the best path forward, but I'm starting to lean in that direction.

gforsyth:
Met with the `cucim` team and we noted that `cucim` is linking against the CUDA driver (will investigate more why that is) -- because of this, we see the import error for `cucim`:

```python
>>> import cucim
dlopen error libcuda.so.1: cannot open shared object file: No such file or directory
 missing cuda symbols while dynamic loading
 cuFile initialization failed
```

This only occurs on CPU runners because `libcuda.so.1` isn't available.

gforsyth:
Confirmed that we now error out if combined stdout+stderr contains any import errors, even if the Python exit code is zero here: https://github.com/rapidsai/integration/actions/runs/24741791136/job/72383206998?pr=837

(That also includes an array of strings to search for in output that will cause us to mark the import test as a failure, we can add to that as needed).

<details>
<summary> Output showing import failure of `cucim` </summary>

```sh
[6](https://github.com/rapidsai/integration/actions/runs/24742914089/job/72387020529?pr=837#step:13:717)
RAPIDS logger » [04/21/26 19:28:20]
┌───────────────────────────────────────┐
|    Standalone import test for cudf    |
└───────────────────────────────────────┘


RAPIDS logger » [04/21/26 19:28:27]
┌──────────────┐
|    Passed    |
└──────────────┘


RAPIDS logger » [04/21/26 19:28:27]
┌────────────────────────────────────────────┐
|    Standalone import test for dask_cudf    |
└────────────────────────────────────────────┘


RAPIDS logger » [04/21/26 19:28:30]
┌──────────────┐
|    Passed    |
└──────────────┘


RAPIDS logger » [04/21/26 19:28:30]
┌───────────────────────────────────────┐
|    Standalone import test for cuml    |
└───────────────────────────────────────┘


RAPIDS logger » [04/21/26 19:28:35]
┌──────────────┐
|    Passed    |
└──────────────┘


RAPIDS logger » [04/21/26 19:28:35]
┌────────────────────────────────────────────┐
|    Standalone import test for pylibraft    |
└────────────────────────────────────────────┘


RAPIDS logger » [04/21/26 19:28:35]
┌──────────────┐
|    Passed    |
└──────────────┘


RAPIDS logger » [04/21/26 19:28:35]
┌────────────────────────────────────────────┐
|    Standalone import test for raft_dask    |
└────────────────────────────────────────────┘


RAPIDS logger » [04/21/26 19:28:35]
┌──────────────┐
|    Passed    |
└──────────────┘


RAPIDS logger » [04/21/26 19:28:35]
┌──────────────────────────────────────────┐
|    Standalone import test for cugraph    |
└──────────────────────────────────────────┘


RAPIDS logger » [04/21/26 19:28:37]
┌──────────────┐
|    Passed    |
└──────────────┘


RAPIDS logger » [04/21/26 19:28:37]
┌─────────────────────────────────────────────┐
|    Standalone import test for nx_cugraph    |
└─────────────────────────────────────────────┘


RAPIDS logger » [04/21/26 19:28:39]
┌──────────────┐
|    Passed    |
└──────────────┘


RAPIDS logger » [04/21/26 19:28:39]
┌────────────────────────────────────────────┐
|    Standalone import test for cuxfilter    |
└────────────────────────────────────────────┘


RAPIDS logger » [04/21/26 19:28:48]
┌──────────────┐
|    Passed    |
└──────────────┘


RAPIDS logger » [04/21/26 19:28:48]
┌────────────────────────────────────────┐
|    Standalone import test for cucim    |
└────────────────────────────────────────┘

dlopen error libcuda.so.1: cannot open shared object file: No such file or directory
 missing cuda symbols while dynamic loading
 cuFile initialization failed

RAPIDS logger » [04/21/26 19:28:48]
┌─────────────────────────────────────────────────────────────────┐
|    Test failed for: cucim (error pattern detected in output)    |
└─────────────────────────────────────────────────────────────────┘


RAPIDS logger » [04/21/26 19:28:48]
┌───────────────────────────────────────┐
|    Standalone import test for cuvs    |
└───────────────────────────────────────┘


RAPIDS logger » [04/21/26 19:28:48]
┌──────────────┐
|    Passed    |
└──────────────┘


RAPIDS logger » [04/21/26 19:28:48]
┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
|    Combined import test for: cudf dask_cudf cuml pylibraft raft_dask cugraph nx_cugraph cuxfilter cucim cuvs    |
└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘

dlopen error libcuda.so.1: cannot open shared object file: No such file or directory
 missing cuda symbols while dynamic loading
 cuFile initialization failed

RAPIDS logger » [04/21/26 19:28:57]
┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
|    Test failed for: cudf dask_cudf cuml pylibraft raft_dask cugraph nx_cugraph cuxfilter cucim cuvs (error pattern detected in output)    |
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘

Error: Process completed with exit code 2.
```

</details>

After skipping `cucim` cpu import tests, since they link against `libcuda.so.1`, all the stable import tests are passing, see https://github.com/rapidsai/integration/actions/runs/24781592917/

I'll go ahead and remove the test runs from `pr.yaml` and keep them in `test.yaml`

ci/stable_install/install_and_test_pip.sh
gforsyth: This is done.

.github/workflows/test.yaml
```diff
@@ -28,3 +28,50 @@ jobs:
...
+      fail-fast: false
+      matrix:
+        arch:
+          - amd64
+          - arm64
+    with:
+      build_type: pull-request
+      node_type: "cpu4"
+      arch: "${{matrix.arch}}"
+      container_image: "rapidsai/citestwheel:cuda12.9.1-ubuntu24.04-py3.14"
```
bdice: I’d like to use the new compute-matrix workflows in shared-workflows so it stays in sync with our testing matrix, perhaps instead of the loop in the bash script. We need coverage of minimum CUDA and Python versions too.

ci/release/update-version.sh
```diff
@@ -94,6 +94,7 @@ function sed_runner() {
 
 sed_runner "/RAPIDS_VER=/ s/[0-9][0-9].[0-9][0-9]/${NEXT_SHORT_TAG}/" ci/conda-pack.sh
 sed_runner "/RAPIDS_VERSION=/ s/[0-9][0-9].[0-9][0-9]/${NEXT_SHORT_TAG}/" ci/test_conda_nightly_env.sh
+sed_runner "/STABLE_RAPIDS_VERSION=/ s/[0-9][0-9].[0-9][0-9]/${NEXT_SHORT_TAG}/" ci/stable_install/install_and_test_*.sh
```
bdice: This needs to lag one release, right? On main, we are testing 26.04.

ci/stable_install/install_and_test_conda.sh
```diff
@@ -0,0 +1,63 @@
...
+
+        rapids-logger "Testing stable version install with Python $PY_VER and CUDA version $CUDA_VERSION"
+
+        # use `-O` to override channels so we don't include `rapidsai-nightly`
+        conda create -n "$envName" -O -c rapidsai -c conda-forge -y \
+          rapids="$STABLE_RAPIDS_VERSION" python="$PY_VER" "cuda-version==${CUDA_VERSION}"
+
+        conda activate "$envName"
+
+        testImports cudf dask_cudf cuml pylibraft raft_dask cugraph nx_cugraph cuxfilter cuvs # cucim
```
bdice: Do we want cudf-polars, rapidsmpf, rmm …? There are many packages not listed.

ci/stable_install/install_and_test_pip.sh
```diff
@@ -0,0 +1,123 @@
...
+
+    for cuda_major_minor in "${CUDA_MINOR_VERSIONS[@]}"; do
+        rapids-logger "Using cuda-toolkit for CUDA ${cuda_major_minor}"
+        PIP_INSTALL_PYPI=(
+            "cudf-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}"
+            "dask-cudf-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}"
+            "cuml-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}"
+            "pylibraft-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}"
+            "raft-dask-${CUDA_SUFFIX}==${STABLE_RAPIDS_VERSION}"
+            "cuda-toolkit[cublas,cufft,curand,cusolver,cusparse,nvcc,nvrtc]${cuda_major_minor}"
```
bdice: I believe you can pin just the metapackage with no extras. The extras should be included in the solve by the other libraries. In fact, that should help ensure we have listed the right extras.

ci/stable_install/install_and_test_pip.sh
```diff
@@ -0,0 +1,123 @@
...
+  esac
+done
+
+function createPyEnv {
+    PY_VER=$1
+    INSTALL_DIR=$2
+    pushd "$INSTALL_DIR"
+
+    rapids-logger "Creating virtualenv for Python $PY_VER"
+    uv venv --python="$PY_VER" --seed
```
bdice, line 62: We’re still testing pip as the solver when the env is created with uv, right? I love uv but want to test vanilla pip here.

bdice (commented):
Please compare this to the proposals in https://github.com/rapidsai/integration/issues/120. I think it’ll close that issue.

For bonus points we should add a job that tests common extras: is stable RAPIDS solvable with PyTorch? See also: https://github.com/rapidsai/integration/issues/794

ci/stable_install/install_and_test_pip.sh
gforsyth: James had a reason he wanted the extras explicitly listed but I can't remember what it was -- happy to go either way

ci/stable_install/install_and_test_pip.sh
gforsyth: Yes, `uv` is just a convenience for bootstrapping python version + virtualenv

ci/release/update-version.sh
gforsyth: good catch

ci/stable_install/install_and_test_conda.sh
gforsyth: I was initially copying what's listed in the basic install docs, but I can add more packages

ci/release/update-version.sh
```diff
@@ -94,6 +99,7 @@ function sed_runner() {
 
 sed_runner "/RAPIDS_VER=/ s/[0-9][0-9].[0-9][0-9]/${NEXT_SHORT_TAG}/" ci/conda-pack.sh
 sed_runner "/RAPIDS_VERSION=/ s/[0-9][0-9].[0-9][0-9]/${NEXT_SHORT_TAG}/" ci/test_conda_nightly_env.sh
+sed_runner "/STABLE_RAPIDS_VERSION=/ s/[0-9][0-9]\.[0-9][0-9]/${CURRENT_SHORT_TAG}/" ci/stable_install/install_and_test_*.sh
```
gforsyth, line 102: when we bump to the next alpha release, set the STABLE version to the "CURRENT_SHORT_TAG" which will be the latest released version

ci/stable_install/bootstrap/pip.sh
```diff
@@ -0,0 +1,21 @@
...
+echo "installing 'uv'"
+curl -LsSf https://astral.sh/uv/install.sh | sh
+echo "done installing 'uv'"
+
+source "$HOME"/.local/bin/env
+
+# TODO: remove me after updating to 26.06
+# this will fail once we pull newer containers that have already moved
+# `pip.conf` to the `/etc/pip.conf` location
+mv /etc/xdg/pip/pip.conf /etc/pip.conf
```
gforsyth, line 15: A small annoyance, but the `mv` will fail once we have newer containers that have the `/etc/pip.conf` location changed and so we'll get the signal that we can remove this small workaround

ci/stable_install/install_and_test_conda.sh
```diff
@@ -0,0 +1,83 @@
...
+  cuml
+  cuvs
+  cuxfilter
+  dask_cudf
+  nx_cugraph
+  pylibraft
+  raft_dask
+  rmm
+)
+testImports RAPIDS_IMPORTS
```
gforsyth, line 80: I'm defining `RAPIDS_IMPORTS` as an array here, but then passing it by name to `testImports` to get around some array expansion shenanigans -- this triggers `shellcheck` into thinking the variable is unused, so I have to disable that check here.

ci/stable_install/install_and_test_pip.sh
```diff
@@ -0,0 +1,139 @@
...
+  cuvs
+  cuxfilter
+  dask_cudf
+  nx_cugraph
+  pylibraft
+  raft_dask
+  rapidsmpf
+  rmm
+)
+testImports RAPIDS_IMPORTS
```
gforsyth, line 134: I'm defining `RAPIDS_IMPORTS` as an array here, but then passing it by name to `testImports` to get around some array expansion shenanigans -- this triggers `shellcheck` into thinking the variable is unused, so I have to disable that check here.

ci/stable_install/install_and_test_pip.sh
```diff
@@ -0,0 +1,139 @@
...
+
+CUDA_SUFFIX="cu${CUDA_VERSION%%.*}"
+
+# TODO: CTK support for cublas on arm is weird around 12.2 -- we can remove this
+# workaround when support for 12.2 is dropped
+if [[ "${CUDA_VERSION}" == 12.2* ]] && [[ "$(uname -m)" == "aarch64" ]]; then
+    CUDA_TOOLKIT_SPEC="cuda-toolkit>=12.2,<12.4"
+else
+    CUDA_TOOLKIT_SPEC="cuda-toolkit==${CUDA_VERSION}.*"
+fi
```
gforsyth, line 74: `cublas` on arm is a little wonky for CTK 12.2.2, so for that combination of architecture and CTK, we need to loosen the bounds a little bit.

gforsyth:
> For bonus points we should add a job that tests common extras: is stable RAPIDS solvable with PyTorch?

I think this would be a good fast-follow add on to this

.github/workflows/pr.yaml
```diff
@@ -15,7 +15,7 @@ jobs:
       - build
       - checks
       - check-nightly-ci
-      - test-conda-nightly-env
+      # - test-conda-nightly-env
```
bdice: I think this is being temporarily commented, please revert if so.

bdice (approved):

.github/workflows/pr.yaml
gforsyth: Good catch, thanks.  Also removed the stable install tests from `pr.yaml`, as we're going to run them as nightlies

ci/stable_install/install_and_test_pip.sh
jameslamb: > _James had a reason he wanted the extras explicitly listed but I can't remember what it was_

Because without them, you can get a solve like:

```text
cuda-toolkit==12.9.1
nvidia-cublas-cu12=={version-from-CUDA-13.0}
```

The solver can choose to work around conflicts by falling back to packages that pinned against `nvidia-{component}` wheels instead of `cuda-toolkit`.

Using the extras prevents that.

ci/stable_install/install_and_test_pip.sh
jameslamb: Yep definitely, I did stuff like this all over RAPIDS in https://github.com/rapidsai/build-planning/issues/256

e.g. https://github.com/rapidsai/cuvs/blob/fc3d6c1647ff512ec8a7e6d5f4ef3880d4064f09/dependencies.yaml#L332-L335

ci/stable_install/install_and_test_pip.sh
jameslamb: That said I'm fine with this how it is, without the extras. No need for a followup.

jameslamb:
Reviewed this long after merging just because I was so involved from the beginning, and then went out on leave before it was finished.

I'm very happy with where we ended up, great work Gil!

PR #839, opened 2026-04-30
feat: add `rapids doctor` smoketest to stable install tests

gforsyth:
xref rapidsai/build-planning#226

Adding a `rapids-cli` smoke test -- this currently clones from upstream `main` for `rapids-cli` -- once there is a more regular release cadence, we can pull from PyPI.

PR #840, opened 2026-04-30
feat: add uv installation, import testing, and smoke tests

gforsyth:
Built on top of #839

Phase 3 of rapidsai/build-planning#226


gforsyth:
Adding packages that are only on `pypi.nvidia.com` changes the index precedence for packages that exist on both indices -- I need to look at this more

Issue #846, opened 2026-05-28
nightly tests broken on `main`: CUDA 13.2 images not found


jameslamb:
## Description

The `test` job has broken on `main` for a while, like this:

>  _Error response from daemon: failed to resolve reference "docker.io/rapidsai/ci-conda:26.04-cuda13.2.0-ubuntu24.04-py3.14": docker.io/rapidsai/ci-conda:26.04-cuda13.2.0-ubuntu24.04-py3.14: not found_

https://github.com/rapidsai/integration/actions/runs/26564394604/job/78255229268

## Benefits of this work

* fixes nightly tests
* unblocks PR CI (which is failing thanks to the `check-nightly-ci` check)

## Approach

This hard-coded `26.04` is the problem:

https://github.com/rapidsai/integration/blob/b98715e91ffe8969f9f9b5c903bffa0db32cbcb0/.github/workflows/test.yaml#L56

We didn't add CUDA 13.2 support until 26.06 (https://github.com/rapidsai/ci-imgs/pull/382).

Not sure of the most future-proof way to fix that though, this needs investigation.

## Notes

N/A

gforsyth:
Ok, so we bump the testing matrix values to reflect our _current_ development, but the stable install testing, by definition, works with the dependencies available in the last release.

I think we probably want to take an approach similar to `ci-imgs` here, where we have a `matrix.yml` file that lists what the current dependency bounds are and then matrix out the jobs from that (so don't use the `compute-matrix` workflow).

For now, I'm going to `matrix_filter` to drop the `13.2` jobs

gforsyth:
Manual test run passed: https://github.com/rapidsai/integration/actions/runs/26587357890

(But I think we can keep this issue open pending a more robust fix)

PR #847, opened 2026-05-28
fix(ci): exclude 13.2.0 from stable install testing on 26.04

gforsyth:
This is a temporary fix for #846 to get the nightly tests passing again.

There is a larger issue around the mismatch between development dependencies in our testing matrix and the dependency matrix at time of release.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -41,6 +41,8 @@ jobs:
     with:
       build_type: nightly
       matrix_name: wheels-test
+      # TODO: add custom dependency matrix for stable install tests
+      matrix_filter: map(select(.CUDA_VER != "13.2.0"))
     permissions:
       contents: read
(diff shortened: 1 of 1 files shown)


jameslamb (approved):

gforsyth:
Going to admin-merge this to fix the nightly ci check

gforsyth:
Blarg -- should I also push this up to the release branch?

jakirkham:
Which release branch? 26.04 already shipped. 26.06 has CUDA 13.2 support.

gforsyth:
> Which release branch? 26.04 already shipped. 26.06 has CUDA 13.2 support.

Yes, but the stable install tests are still pinned to 26.04 images, since they test the latest stable release. 
I think what we should probably actually do is just ignore those tests on the `release` branch, since they only become viable AFTER a release is cut, not while it's in progress.

PR #852, opened 2026-06-09
feat(stable install): bump testing version to latest stable release

gforsyth:
Now that the 26.06 release is out, we can bump the version we test nightly in the stable install tests.

This needs better automation overall.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -42,6 +42,9 @@ jobs:
     with:
       build_type: nightly
       matrix_name: wheels-test
+      # TODO: add custom dependency matrix for stable install tests
+      # Proactively filtering out 13.3 jobs which won't have docker images for the 26.06 release
+      matrix_filter: map(select(.CUDA_VER != "13.3.0"))
     permissions:
       contents: read
   test-stable-install-pip:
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yaml
```diff
@@ -42,8 +42,6 @@ jobs:
     with:
       build_type: nightly
       matrix_name: wheels-test
-      # TODO: add custom dependency matrix for stable install tests
-      matrix_filter: map(select(.CUDA_VER != "13.2.0"))
```
gforsyth, line 46: Shouldn't need these matrix filters anymore now that we're on 26.06 -- but we'll have to add a filter back here when we add CUDA 13.3 support

.github/workflows/test.yaml
```diff
@@ -43,7 +43,8 @@ jobs:
       build_type: nightly
       matrix_name: wheels-test
       # TODO: add custom dependency matrix for stable install tests
-      matrix_filter: map(select(.CUDA_VER != "13.2.0"))
+      # Proactively filtering out 13.3 jobs which won't have docker images for the 26.06 release
+      matrix_filter: map(select(.CUDA_VER != "13.3.0"))
```
gforsyth, line 47: Proactively filtering out 13.3.0 since that cuda version won't have docker iamges built for the 26.06 release

jameslamb (approved):

.github/workflows/test.yaml
jameslamb: yes this is needed, this same change just got made in #853

jameslamb:
ha

PR #854, opened 2026-06-23
Use rapids-cli latest release

ncclementi:
We recently release rapids-cli 0.2.0 so I thought this could be updated here. 

For the record we do have teh release in pypi and in the rapidsai channel for conda. 

I think I cover all the places, but let me know if I missed something. I wasn't quite sure how or if it's possible to to test this locally, though. 

cc; @gforsyth since you put up this one initially https://github.com/rapidsai/integration/pull/839


gforsyth (approved):

gforsyth:
Thanks @ncclementi !

This looks correct to me.

PR #865, opened 2026-08-04
Allow UCX 1.20 in RAPIDS metapackage

jakirkham:
UCX 1.20 has been out for a while and is already tested in RAPIDS. However the metapackage wasn't allowing UCX 1.20 to be installed with the metapackage. Relax the pinning to allow UCX 1.20.


jameslamb (approved):
This is fine, given that `ucxx` is tested against this version.

We can squeeze this in.

jakirkham:
Thanks James! 🙏

PR #868, opened 2026-09-01
Remove RAPIDS XGBoost Conda package

jakirkham:
Now that the RAPIDS XGBoost package built with RMM is deprecated, drop the `rapids-xgboost` metapackage. Going forward the conda-forge `xgboost` package should be used instead.

As part of this require XGBoost 3.4.0+, which is only available via conda-forge (not rapidsai).


jakirkham:
Putting an initial pass up for feedback

Requested reviews from a few people

Further changes may still be needed

Marked as `DO NOT MERGE` while we iterate

bdice (approved):

bdice (dismissed):
Additional removals needed:

https://github.com/rapidsai/integration/blob/98a78aabbef4a40ede45a9eb686351c6c544fb63/conda/recipes/README.md?plain=1#L27

https://github.com/rapidsai/integration/blob/98a78aabbef4a40ede45a9eb686351c6c544fb63/ci/build_python.sh#L19

https://github.com/rapidsai/integration/blob/98a78aabbef4a40ede45a9eb686351c6c544fb63/ci/check_conda_nightly_env.py#L12

trivialfis (approved):

gforsyth:
I think the nightly env is going to inherently fail until we've published a nightly that _doesn't_ rely on `rapids-xgboost`

jakirkham:
Thanks Bradley and Gil! 🙏 

True. Though it is more than that

Looking at [one of the CI logs]( https://github.com/rapidsai/integration/actions/runs/34487358320/job/102905476010?pr=868#step:11:59 ), we see the following error:

```
nothing provides rapids-xgboost 26.10.* needed by rapids-26.10.00a6-cuda13_260904070554_98a78aab
```

IOW [the environment creation test uses the `rapidsai-nightly` channel]( https://github.com/rapidsai/integration/blob/b77b371c72b8d6a143eac12b7343f2b03c600174/ci/test_conda_nightly_env.sh#L19 ) and thus fails to pickup the package built in the PR. This will always fail. So we need to change this test to prefer the package built in the PR instead

Since this test can run on `main` where no PR package build is relevant, it will still need to use the `rapidsai-nightly` package as a fallback. We should just prefer the PR build cache first

conda/recipes/versions.yaml
```diff
@@ -14,3 +11,5 @@ nvtx_version:
   - '>=0.2.1,<0.3'
 ucx_version:
   - '>=1.15.0,<1.21.0'
+xgboost_version:
+  - '>=3.4.0,<3.5.0'
```
bdice, line 15: We have a problem here that hasn't been discussed yet. We will have to do one of the following for 26.10:

Pick a version bound of `xgboost` that supports Python 3.11, like `>=3.2.0,<3.5.0`. @chyunsu3 explored a backport of 3.11 to xgboost 3.3.0: https://github.com/dmlc/xgboost/pull/12374 We backported that to our fork in https://github.com/rapidsai/xgboost-feedstock/pull/146 but conda-forge packages of xgboost 3.3.0 and newer do not support Python 3.11.

Alternatively we can drop `xgboost` from the `rapids` metapackage and require users to install it themselves in environments with a supported Python version. We should _not_ ship multiple metapackages that differ in their contents by Python version (e.g. Python 3.11 without `xgboost` and newer Python versions with `xgboost`).

cc: @mgrauer @trivialfis for vis

bdice:
The artifact changes in 3b33ff3 are working: CI now selects the PR-built `rapids` package. The [current CI run](https://github.com/rapidsai/integration/actions/runs/34706817093) has two remaining failure causes:

- **4 Python 3.11 jobs:** XGBoost 3.4 requires Python >=3.12. See https://github.com/rapidsai/integration/pull/868#discussion_r3996775533.
- **12 Python 3.12–3.14 jobs:** environment solves succeed, but the freshness check rejects September 8 builds of `nx-cugraph` and `dask-cuda`.

Fresh `release/26.10` packages have been built and uploaded, but their versions sort below the older packages:

| Package | September 8 version selected by CI | Fresh September 12 version |
|---|---|---|
| `nx-cugraph` | `26.10.00a11` | `26.10.00a8` |
| `dask-cuda` | `26.10.00a17` | `26.10.00a14` |

For example, [this nx-cugraph build](https://github.com/rapidsai/nx-cugraph/actions/runs/34674787224) successfully uploaded the fresh `a8` package. Both versions satisfy `26.10.*`, so conda prefers the higher version despite its older build date. The [cleanup policy](https://github.com/rapidsai/remove-old-packages/blob/eb9a3134d24a516429d4908be18825f7a59b34e4/config.yaml) retains at least 12 files per version family and architecture, so these packages remain protected even though they exceed the three-day freshness threshold.

@rapidsai/ops could you please remove these outdated distributions from `rapidsai-nightly` so conda can select the fresh release-branch builds? We also need to figure out how to prevent this alpha-numbering issue from reoccurring, but I don't know the burndown process well enough to advise on that.

- `noarch/nx-cugraph-26.10.00a11-py_260908205229_c0273373.conda`
- `noarch/dask-cuda-26.10.00a17-py314_260908204825_5c7162a8.conda`

msarahan:
I moved those 2 packages to a "broken" label.

msarahan:
The xgboost switcheroo is done. New build number 1 packages are in place. Merging this PR so that integration can pass.

msarahan (approved):

